Issue #42, opened 2020-12-14
Add option to use full RAM/VRAM for pool_patch


Trypticon8:
After running the game with the latest plugin update and the mem pool fix enabled, I noticed a (slight) performance decrease and a general decrease in stability for my framerates as shown below. I had been using the edited .csv config sheet with my full amount of memory for both CPU and GPU as it had not caused any instability in the game for me, nor any PC crashes.

- With [PerfOverhaul](https://i.imgur.com/yyYbjaH.jpeg)'s memory fix 

- Mem fix disabled, using edited config set to my full amount of RAM/VRAM [here](https://i.imgur.com/CgYGJzo.jpeg)

Took a peek at the code for "pool_patch.cpp" and noticed on Line 60 that the patch subtracts 4GB from available memory as a crash precaution : `aSize = (gigsInstalled - 4) * kScaler;`. I'm not versed at all in C++ so I haven't tried to make a full edit of the file myself. I'd like to have a toggle that enables me to use the full amount of memory available in my system (with an added disclaimer should there be any crashing issues of course) if possible, as I'm currently having to disable the mem fix in the plugin and use the edited config instead. 

Also quick thanks for making this plugin, makes playing the game properly _far_ more convenient.

yamashi:
Will think about it :)

yobson1:
Perhaps a memory percentage option would be a better solution.

Pogogo007:
Im actually gonna try making this right now. No promises tho

PR #47, opened 2020-12-14
added true memory

Pogogo007:
Added #42

### DIFF
--- a/cyberpunk_patch/src/pool_patch.cpp
+++ b/cyberpunk_patch/src/pool_patch.cpp
@@ -3,6 +3,7 @@
 #include <mhook-lib/mhook.h>
 #include <d3d11.h>
 #include <atlbase.h>
+#include <Options.h>
 
 using TRegisterPoolOptions = void(void*, const char*, uint64_t);
 TRegisterPoolOptions* RealRegisterPoolOptions = nullptr;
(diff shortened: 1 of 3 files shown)


Pogogo007:
Im not an adept at c++ so a little cleanup would be nice

maximegmd (changes requested):
Pass options to the function, don't create one like that, this can cause crashes.

Pogogo007:
on it

Pogogo007:
No idea if i did it correctly

cyberpunk_patch/src/pool_patch.cpp
```diff
@@ -57,11 +57,11 @@ void RegisterPoolOptions(void* apThis, const char* acpName, uint64_t aSize)
         if (statex.ullTotalPhys)
         {
             const auto gigsInstalled = statex.ullTotalPhys / kScaler;
-            HMODULE AModule;
-            Options options(AModule);
+            Options options(mod);
```
PhantomGamers, line 60: This is what should be passed to the function

maximegmd (commented):
Oh right this is called in the hook so it cant be passed this way, I guess it's time to make Options a singleton...

Pogogo007:
> Oh right this is called in the hook so it cant be passed this way, I guess it's time to make Options a singleton...

no idea what that means but ill let ya do your thing

yamashi:
Appreciate you trying, please do look at the solution I will post for this tomorrow, I will explain the commit why it has to be done that way

Issue #81, opened 2020-12-16
Setting up the mod with ProtonTricks not needed on gog version of the game


ArturVRSampaio:
actually the proton tricks don't work at all if you don't have the game on steam... on gog version i just updated my x64 folder

PR #82, opened 2020-12-16
Added Winetricks instructions for GOG version users.

PoorPocketsMcNewHold:
Would adding pictures to the installation help? I can provide those.
Address #81


ArturVRSampaio:
> u

well, not that much to see, i just updated my x64... just like the normal install guide

SpartanDraco:
It is the same installation but GOG, I have it from GOG and I have installed it as it says it but being logical that the part of where it is installed changes the rest is the same

PR #149, opened 2020-12-19
Made the console key configurable

### LINKED ISSUE #133: Change console toggle key
End apparently conflicts with Reshade, we should use another default and it would be great to be able to configure the key.

Kyle873:
The end-user will need to refer to the [virtual key codes](https://docs.microsoft.com/en-us/windows/win32/inputdev/virtual-key-codes) if they wish to change it. This should resolve #133.


DartPower:
Wow, I didn't expect to see you here o__O
P.S. Thanks for best project DoomRPG ^__^

PR #156, opened 2020-12-20
Console improvements.

gibbed:
- Automatically resize console output area to match window.
- Copy selected console output line to current input.
- Checkbox to enable auto scrolling of output area.
- Improved parsing of commands.
- More sanity checking on executing command function.
- Don't require flag 0 (global?) on command functions.
- Don't block all window messages, block specific ones instead.


yamashi (approved):

yamashi:
Thanks a lot for the PR!

PR #191, opened 2020-12-21
Add sanity check for console parameters

WopsS:
Right now we cannot handle other types of parameters, passing an incorrect type would likely cause the game to crash.


yamashi:
Poggies that's awesome

PR #192, opened 2020-12-21
Allow Lua to access GameOptions / INI settings

emoose:
- Removed GamePropertyString, use REDString instead
- Renamed GameProperty to GameOption for uniformity & moved GameOption code to reverse\GameOptions.cpp/.h
- Add GameOptions struct to hold the list of GameOption pointers
- Add GameOption::Get/Set functions
- Made the GameOptionInitHook code always get run without needing 'dump_game_options', so we can capture all the GameOption pointers

Also added the following to GameOptions & bound them to Lua, can be accessed via:
- `GameOptions.Print(category, name)` - prints option value to console
- `GameOptions.List(category)` - prints option name/values to console, optional category to filter to
- `GameOptions.Get(category, name)` - returns option value as string
- `GameOptions.GetBool(category, name)` - returns option value as bool
- `GameOptions.GetInt(category, name)` - as int
- `GameOptions.GetFloat(category, name)` - as float
- `GameOptions.Set(category, name, valueStr)` - valueStr will attempt to be converted to options type
- `GameOptions.SetBool(category, name, valueBool)`
- `GameOptions.SetInt(category, name, valueInt)`
- `GameOptions.SetFloat(category, name, valueFloat)`
- `GameOptions.Toggle(category, name)` - inverts value of a boolean
- `GameOptions.Dump()` - dumps current option values to log file

(the different SetInt/SetFloat/etc functions are to allow Lua int/floats to be used without needing conversion, not 100% sure if this is necessary though)

Set functions will also respect the options min/max too if the game defined any. ATM `String` GameOptions can't be edited since I haven't looked into finding the REDString::Set function in the game yet, not many interesting options use that anyway.

Anyway with this you can edit the INI settings at runtime and usually see the results instantly, mostly wanted this so I could look into the "DRS" (Dynamic Resolution Scaling?) options, seems with all resolution scaling stuff disabled in the ingame settings this still has some effect on things, which can be seen by running the following a few times:
`GameOptions.Toggle("DRS", "Enable")`

Still not sure if it is resolution related though, or maybe just something to do with lights...

(A list of all the INI settings that you can use with this is at https://www.nexusmods.com/cyberpunk2077/mods/193)


emoose:
Oops, forgot to bind the Toggle function, 1 sec

emoose:
Renamed GameOptions::Get to Print & added Get/GetBool/GetInt/GetFloat funcs that return the value instead of printing it directly.

Also in future I think it'd be nicer if we could handle the GameOption structs themselves in Lua, like
```
opt = GameOptions("DRS", "Enable")
opt.SetBool(False)
```
Haven't been able to figure out exposing the GameOption* to Lua properly yet though, only get crashes trying that method atm... might be related to the virtual destructor, not sure.

yamashi:
If I understand correctly this isn't ready to be merged right?

emoose:
It should be fine to merge, not sure if the stuff I mentioned in last post could actually be done or not, maybe something to look into later on in the future.

Merge conflict should hopefully be fixed now too.

yamashi:
Oh btw you should join https://discord.gg/SmVRdntx
Ping me when you are there so I give you roles to access the dev section

PR #241, opened 2020-12-22
Added custom "load" function

arwent:
Can load multiples file at once

I still offer this command because "dofile" is working, but for those who would be interested to load multiple files or if you find that this feature is cool, not sure it'll be accepted


yamashi:
I don't really see the point of this, something useful would be to have a way to register functions for events.

Such as 
```lua
s = CreateEntity()
s:RegisterForEvent("update", function(self, delta)
    print(delta)
end)
```

PR #255, opened 2020-12-23
TweakDBID lookup.

gibbed:


gibbed:
`RED4ext::REDreverse::CString::ToString` needs const-ness added.

Issue #280, opened 2020-12-24
Game crash when open console.


lukenYang:
cyber_engine_tweaks.log
[2020-12-24 23:16:12.188] [info] Game version 1.06
[2020-12-24 23:16:12.223] [info] 	AMD SMT Patch: success
**[2020-12-24 23:16:12.223] [warning] 	Spectre patch: failed, unknown version**
**[2020-12-24 23:16:12.236] [warning] 	Virtual Input Patch: failed**
[2020-12-24 23:16:12.280] [info] 	Start screen patch: success
[2020-12-24 23:16:12.283] [info] 	Disable intro movies patch: success
[2020-12-24 23:16:12.358] [info] 	Hidden options hook: success
[2020-12-24 23:16:12.393] [info] 	Hook mouse clip function!
[2020-12-24 23:16:12.421] [info] 	Log function hook complete!
[2020-12-24 23:16:12.450] [info] 	LogChannel function hook complete!
[2020-12-24 23:16:12.524] [info] 	TDBID::ctor function hook complete!
[2020-12-24 23:16:12.599] [info] 	TDBID::ctor[CString] function hook complete!
[2020-12-24 23:16:12.673] [info] 	TDBID::ctor[Derive] function hook complete!
[2020-12-24 23:16:12.747] [info] 	TDBID::ctor[Unknown] function hook complete!
[2020-12-24 23:16:12.786] [info] 	TDBID::ToStringDEBUG function hook complete!
[2020-12-24 23:16:13.240] [info] 	D3D12 hook complete

stacktrace.txt
Error reason: Unhandled exception
Expression: EXCEPTION_ACCESS_VIOLATION (0xC0000005)
Message: The thread attempted to read inaccessible data at 0x0.
<Unknown>(0)

report.txt
Version=1
EventType=REDEngineErrorReport
EventTime=0
ReportIdentifier=ffffffff-ffffffff-00000001-00000000
Sig[0].Name=StackHash
Sig[0].Value=0x0000000000000000
Sig[1].Name=ErrorReason
Sig[1].Value=Unhandled exception
Sig[2].Name=InternalVersion
Sig[2].Value=3.0.3235023  P4CL: 4177747  Stream: //R6.Root/R6.Patch0Hotfix2 
DynamicSig[1].Name=OS Version
DynamicSig[1].Value=10.0.18363.2.0.0.256.1

psychomax13:
The same. Just crashed on 1.06.

Here is the log.

[...]

Issue #267, opened 2020-12-23
[1.06] Openning console results in CTD with RTSS running


WSSDude420:
Latest version of the mod seems to work fine with one exception...

Console is not working... That is the only thing I found out wrong with latest version.

WSSDude420:
Sorry, I can see that in the mean time, you started committing fixes... 

Closing this issue, sorry for disturbance 🙏

WSSDude420:
Actually reopening - latest fixes didn't help me...

Still crashes to desktop when I try to open console on latest version.

yamashi:
Can't repro, please send log file and list of stuff you have installed.

WSSDude420:
Ok, found what's up...

RivaTuner Server Statistics crashes it. This is new behaviour though, so still should be noted I believe.
With 1.05, having RTSS running was no issue, openning console would just not render it (fixed though with https://github.com/yamashi/CyberEngineTweaks/issues/245#issuecomment-750009640 so this was ultimately non-issue)

Now, having RTSS on causes CTD when openning console.
Completely shutting down RTSS helps this issue, console works fine.

Log attached. It is reproducible 100% of the time for me.
[Cyberpunk2077-20201223-235009-17280-2248.zip](https://github.com/yamashi/CyberEngineTweaks/files/5737404/Cyberpunk2077-20201223-235009-17280-2248.zip)

WSSDude420:
Ok, it may still somehow be caused by order...

Fix for 1.05 is not working for 1.06 it seems, RTSS seems to inject first for some reason now even with it applied...
This causes the crash. Previously, it would just not render the console in this case as said. Now it crashes.

Just confirmed this by few tests. 
Launching CP77 with RTSS completely turned off allows me to use console without crashes. Starting RTSS afterwards produces wanted result (running game, RTSS on, console working).

I use version 7.2.3 of RTSS... Should have probably said that sooner.

WSSDude420:
Should be fixed to an extent with upcoming release now.

Note though that you should follow this comment for RTSS to work reliably (otherwise, console works half the time and other half is not rendering, depending on which hook was faster...)
#245 (comment)

Working on revising whole DX12 hooking procedure now, so hopefully, we will have some more robust solution in place SOON™ which would not require similar stunts from user.

Issue #260, opened 2020-12-23
Crash on opening console


Gyunbie:
I found the conflicting mod. It's Autonomous ReShade. You have to choose one of them until a fix comes.

WSSDude420:
Did you test latest release here? Don't know now if this was posted before or after it now...

Seems to work just fine for me with latest release here, that is why I'm surprised a bit.

PR #285, opened 2020-12-25
Implemented better method for command queue retrieval

WSSDude420:
This method uses offset from swapchain pointer. This is much more reliable, as other DX12 hooks may interfere with previous implementation.

Helps prevent crashing with some applications requiring DX12 hook.

Partially helps #280 and #267 . CTD is not happening anymore, but we still have an issue with rendering it seems because of order of hooking with potential other applications it seems.

Functionality with this does not break from what I tested regarding running solely this DLL.

This also helps #267 specifically in one more way than not crashing - it allows to utilize fix from https://github.com/yamashi/CyberEngineTweaks/issues/245#issuecomment-750009640 to use RTSS with this mod, like it was back in 1.04 and 1.05.

I want to have a look into that ordering or whatever may be a problem next.

### DIFF
--- a/src/overlay/Overlay.h
+++ b/src/overlay/Overlay.h
@@ -67,7 +67,6 @@ protected:
 	void DrawImgui(IDXGISwapChain3* apSwapChain);
 
 	static long PresentD3D12(IDXGISwapChain3* pSwapChain, UINT SyncInterval, UINT Flags);
-	static void ExecuteCommandListsD3D12(ID3D12CommandQueue* apCommandQueue, UINT NumCommandLists, ID3D12CommandList* const* ppCommandLists);
 	static BOOL SetMousePosition(void* apThis, HWND Wnd, long X, long Y);
 	static BOOL ClipToCenter(CGameEngine::UnkC0* apThis);
 	static LRESULT APIENTRY WndProc(HWND hwnd, UINT uMsg, WPARAM wParam, LPARAM lParam);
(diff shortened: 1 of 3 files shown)


WSSDude420:
One note I'd like to add is that the problem with CTDs was m_pCommandQueue being nullptr.

Removed hook in this commit was not working properly in case of other DX12 hook running it looked like for one reason or another...

This is method I found should be more reliable and the way kiero handles it seems to be as proper as it gets.

WSSDude420:
Looked also into issue #260 now, seems like there will be some other issue regarding command queue still... As we are still crashing with Reshade with this fix...

Same error as with RTSS - m_pCommandQueue results in nullptr, even with this change...

So while it may be better that we have this now inside initialize, there is still some weird issue going on it seems...

yamashi:
So you are basically reverting the reshade fix... doesnt make any sense if I stopped doing it like this there is a good reason.

WSSDude420:
I didnt know it was otherwise before, didnt check history.

This explaqins why it worked before and stopped after...

So, you are effectively telling here now that you fix single DLL effectively and screw others...
And you do not see a good reason for revert...

This is how it should be done as this is more reliable. Reshade must be broken in some other way.
As said, there seems to be generally some problem with when this hooks... It needs to be the first to hook it seems, which is all there seems to it...

yamashi:
The command queue is nullptr when using this method with reshade, so a solution could be to use both methods, if it's not nullptr, use it, otherwise use the hook.

WSSDude420:
Worth a check I guess... Will give that a try and report back if you want. I kind of think I signed with the devil already anyway trying to fix it somehow.

WSSDude420:
You were right in this one - seems having both fixes issues with both, Reshade and RTSS.

New pull request in a moment (as you closed this one already).

Found out some things worth checking though, will write about it more in next PR.

PR #290, opened 2020-12-25
DX12 hook CTD fix for RTSS, not breaking Reshade (and probably some other as well)

WSSDude420:
Following #285, this should be better fix for issues with both, RTSS and Reshade working now.

Seems this is all that was needed to do...

Few problems still though worth revising. Was digging around how Reshade does this hooking procedure as a whole.

Seems we may have an issue on Windows 7, as DX12 version for Windows 7 apparently does not use swapchains... So that is one no-no we have not properly handled yet I believe, as we now rely on swapchain being present.

Another thing is that getting command queue this way seems unsafe in general after looking into said Reshade. They seem to have much smarter way of handling this implemented, which seems to work even for Windows 7 (is not using swapchain at all for this).

We also use raw pointers, not properly increasing reference counts and such in ComPtr objects we retrieve pointers from. Reshade does properly increase ref. counts and such. We don't have an issue with this currently it seems, but we may have some time later. 
Reshade has implemented tiny ComPtr interface which manages this for them partially (seems to be bit of a rip-off of ATLs CComPtr, so I guess we could use that as we already depend on ATL). Probably should correct this too in near future.

Just letting you know about my findings. I'd like to work on this a bit, hopefully revising all the things seemingly wrong with it to some extent so we have more robust solution here.

### DIFF
--- a/src/overlay/Overlay.h
+++ b/src/overlay/Overlay.h
@@ -67,6 +67,7 @@ protected:
 	void DrawImgui(IDXGISwapChain3* apSwapChain);
 
 	static long PresentD3D12(IDXGISwapChain3* pSwapChain, UINT SyncInterval, UINT Flags);
+	static void ExecuteCommandListsD3D12(ID3D12CommandQueue* apCommandQueue, UINT NumCommandLists, ID3D12CommandList* const* ppCommandLists);
 	static BOOL SetMousePosition(void* apThis, HWND Wnd, long X, long Y);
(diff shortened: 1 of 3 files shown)


yamashi:
Dx12 on windows doesn't exist afaik, they use dx11 to emulate it, that's probably why the swapchain is not working correctly.

There are a lot of issues with the dx12 code indeed, it was just something hacked together quickly to get an overlay, I appreciate the effort to fix this!

PR #297, opened 2020-12-26
Update inject/uninject ,Enables Gamepad/Keyboard navigation

KennnyTroll:
1: Update to have files outside the game
cyber_engine_tweaks.asi maybe uninjected and injected many times
deinjected = NumPad -
2: Control the interface with the joystick and the keyboard

### DIFF
--- a/Ze_Open_cmd_In_This_Dir.bat
+++ b/Ze_Open_cmd_In_This_Dir.bat
@@ -1,0 +1,2 @@
+cd /d %~dp0
+cmd.exe
--- a/Ze_Tuto_Build.txt
+++ b/Ze_Tuto_Build.txt
@@ -1,0 +1,28 @@
+0: look 
+https://xmake.io/#/plugin/builtin_plugins
+
+1: run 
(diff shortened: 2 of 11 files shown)


yamashi:
I don't really understand the point of this PR.

Why would we want to uninject? Why are most changes just moving code around?

Why are there random files in french?

Why remove const/noexcept on functions that were const noexcept?

Why are some variables suddenly static even though they are members of a singleton?

Why did Initialize become __stdcall out of nowhere and returns booleans?

Why rename kiero functions into the names of minihook ?

I am sorry but this PR makes no sense to me and looks very confused.

KennnyTroll:
Why French, a typo & googletrans too slow!
Especially impossible to update (it's been 1 year since I no used git ...)
for errors, I'm not a professional coder ...
I just wanted to share what I did and give some ideas .
This is mainly used to run my game once a day and not every time I make a change in my code, I no longer need skip_start_menu & disable_intro_movies ...
And I play with a controller.

Issue #134, opened 2020-12-18
Windows 7 support


yamashi:
Apparently people are still using windows 7 and play directx12 games with a conversion layer...

Currently the overlay hook is based on real directx12, a directx11 hook needs to be developed and used when on windows 7.

I don't plan on implementing it, as it requires installing windows 7 and spending time I don't have on a very small player base.

If you want to give it a shot please do.

asdf2345:
This would take an hour to install, and probably another hour to get all the drivers installed, but it won't take that long, and you'll have a better OS in the end.

https://www.reddit.com/r/LegacyWindows/wiki/index/windows-7/windows-7-iso-with-integrated-updates

r4m0n:
Maybe a config option to disable the overlay then? Should be a lot easier to implement and still keep the general optimization options.

Squall-Leonhart:
> Apparently people are still using windows 7 and play directx12 games with a conversion layer...

12on7 is a complete Dx12 implementation to the equivalent of that included with Windows 10 1809, but limited to the capabilities of DXGI 1.1 where swap chains are concerned.

rrrudd:
Please add option to disable overlay like r4m0n suggested.

Megalomaniak:
I think the issue is not DX12, people can just copy the dlls from the dx12on7 folder to the above folder with the game exe in it so it will be found, but XInput1_4 is what isn't available on windows 7. Could try using/looking for 1_3 for it I think, not sure if that helps tho.

r4m0n:
> I think the issue is not DX12, people can just copy the dlls from the dx12on7 folder to the above folder with the game exe in it so it will be found, but XInput1_4 is what isn't available on windows 7. Could try using/looking for 1_3 for it I think, not sure if that helps tho.

I've tried disabling the xinput option and grabbing a 1.4 xinput DLL. it stops complaining about missing DLLs, but gives a couple errors 193 and doesn't work (game loads normally).

puchev:
> I've tried disabling the xinput option and grabbing a 1.4 xinput DLL. it stops complaining about missing DLLs, but gives a couple errors 193 and doesn't work (game loads normally).
Copying both xinput1_3.dll and d3d12on7/d3d12.dll to plugins/ and renaming the former to xinput1_4.dll should allow the plugin to work on 7 (except for the overlay obviously). Or at least it did, I have yet to update to 1.05 and my plugin is like 2 days old.

rrrudd:
[...]

PR #301, opened 2020-12-26
[imgui/kiero] Dynamically import d3d12.dll

Mattiwatti:
This is required for CET to be able to load on Windows 7 without loading duplicate versions of d3d12.dll. See [this comment on #134](https://github.com/yamashi/CyberEngineTweaks/issues/134#issuecomment-751355427) for the reasoning behind this change.

NB: this does not actually meaningfully improve the D3D12 situation on Windows 7, it only prevents certain issues that would otherwise occur due to DLL duplication if a D3D12On7 implementation were added in the future.


yamashi:
Looks good to me, tested on Windows 10 and everything worked as expected (did not test ReShade/RivaTuner)

PR #303, opened 2020-12-26
[mimalloc] Use RtlGenRandom instead of BCryptGenRandom to init PRNG

Mattiwatti:
This fixes an access violation on Windows 7 during Mimalloc's PRNG initialization (manifests as "error 998" from the ASI loader). See [this comment](https://github.com/yamashi/CyberEngineTweaks/issues/134#issuecomment-751355244) on #134 for details.

Thanks to @waruqi for the help with xmake.

### DIFF
--- a/xmake.lua
+++ b/xmake.lua
@@ -1,5 +1,6 @@
 set_languages("cxx17")
 
+add_requires("mimalloc", {configs = {cxflags = "-DMI_USE_RTLGENRANDOM"}})
 add_requires("spdlog", "nlohmann_json", "minhook", "imgui", "sol2", "tiltedcore", { external = false })
 
 add_rules("mode.debug", "mode.release")


yamashi:
I created a PR to xmake-repo https://github.com/xmake-io/xmake-repo/pull/227

Once it's merged we should use that option instead

Mattiwatti:
Yeah, that is the better solution definitely.

waruqi:
> I created a PR to xmake-repo [xmake-io/xmake-repo#227](https://github.com/xmake-io/xmake-repo/pull/227)
> 
> Once it's merged we should use that option instead

I have merged it. https://github.com/xmake-io/xmake-repo/pull/229

yamashi:
Thanks @waruqi :)

sumyunggai:
Thanks for chasing this down @Mattiwatti, and thanks for supporting Win7 with this fix @yamashi

PR #304, opened 2020-12-26
Overlay refactor

WSSDude420:
Refactored the way we hook a bit. 

Also added some better logging for good measure, so we know sooner what is wrong in regards to overlay hook.

All that should be left to do regarding the overlay is to make it compatible with Windows 7, because as of right now, we are not compatible with it.


src/overlay/Overlay_Hooks.cpp
```diff
@@ -166,15 +150,13 @@ void Overlay::EarlyHooks(Image* apImage)
...
-    {
-        Get().InitializeD3D12(pSwapChain);
-    });
-
-    Get().Render(pSwapChain);
-
-    return Get().m_realPresentD3D12(pSwapChain, SyncInterval, Flags);
+    auto& overlay = Get();
+    
+    if (overlay.InitializeD3D12(pSwapChain)) {
```
yamashi, line 155: line skip

src/overlay/Overlay_D3D12.cpp
```diff
@@ -115,23 +214,23 @@ void Overlay::Render(IDXGISwapChain3* pSwapChain)
     DrawImgui(pSwapChain);
 
     const auto bufferIndex = pSwapChain->GetCurrentBackBufferIndex();
-
-    auto& currentFrameContext = m_frameContexts[bufferIndex];
-    currentFrameContext.CommandAllocator->Reset();
+    auto& frameContext = m_frameContexts[bufferIndex];
+    m_pd3dCommandAllocator->Reset();
```
yamashi, line 215: I remember using the same allocator for every frame would make the overlay randomly working or not, are you sure about that?

src/overlay/Overlay_D3D12.cpp
```diff
@@ -29,82 +28,182 @@ BOOL CALLBACK EnumWindowsProcMy(HWND hwnd, LPARAM lParam)
...
+        m_pd3dSrvDescHeap->GetCPUDescriptorHandleForHeapStart(),
+        m_pd3dSrvDescHeap->GetGPUDescriptorHandleForHeapStart()))
+    {
+        spdlog::error("\tOverlay::InitializeD3D12(pSwapChain={0}) - ImGui_ImplDX12_Init call failed!", reinterpret_cast<void*>(pSwapChain));
+        ImGui_ImplWin32_Shutdown();
+        ImGui::DestroyContext();
+        return reset(this);
+    }
+
+    if (!ImGui_ImplDX12_CreateDeviceObjects()) {
```
yamashi, line 186: line skip

src/overlay/Overlay_D3D12.cpp
```diff
@@ -29,82 +28,182 @@ BOOL CALLBACK EnumWindowsProcMy(HWND hwnd, LPARAM lParam)
...
+        overlay->m_pd3d12Device = nullptr;
+        overlay->m_pd3dRtvDescHeap = nullptr;
+        overlay->m_pd3dSrvDescHeap = nullptr;
+	    overlay->m_pd3dCommandAllocator = nullptr;
+        overlay->m_pd3dCommandList = nullptr;
+        return false;
+    };
+
+    // Window hook (done once)
+    if (m_hWnd == nullptr) {
```
yamashi, line 67: line skip

src/overlay/Overlay_D3D12.cpp
```diff
@@ -29,82 +28,182 @@ BOOL CALLBACK EnumWindowsProcMy(HWND hwnd, LPARAM lParam)
     return TRUE;
 }
 
-void Overlay::InitializeD3D12(IDXGISwapChain3* pSwapChain)
+bool Overlay::InitializeD3D12(IDXGISwapChain3* pSwapChain)
 {
-    EnumWindows(EnumWindowsProcMy, reinterpret_cast<LPARAM>(&m_hwnd));
+    static auto checkCmdQueue = [](Overlay* overlay) {
```
yamashi, line 33: line skip

yamashi: also maybe it should be its own function and not a lambda?

src/overlay/Overlay_D3D12.cpp
```diff
@@ -29,82 +28,182 @@ BOOL CALLBACK EnumWindowsProcMy(HWND hwnd, LPARAM lParam)
...
+                }
+                return true;
+            }
+            spdlog::warn("\tOverlay::InitializeD3D12(pSwapChain={0}) - swap chain is missing command queue!",  reinterpret_cast<void*>(*(&overlay->m_pdxgiSwapChain)));
+            return false;
+        }
+        return true;
+    };
+
+    static auto reset = [](Overlay* overlay) {
```
yamashi, line 55: line skip + same as above

src/overlay/Overlay_D3D12.cpp
```diff
@@ -29,82 +28,182 @@ BOOL CALLBACK EnumWindowsProcMy(HWND hwnd, LPARAM lParam)
...
+        overlay->m_frameContexts.clear();
+        overlay->m_pdxgiSwapChain = nullptr;
+        overlay->m_pd3d12Device = nullptr;
+        overlay->m_pd3dRtvDescHeap = nullptr;
+        overlay->m_pd3dSrvDescHeap = nullptr;
+	    overlay->m_pd3dCommandAllocator = nullptr;
```
yamashi, line 61: indentation

src/dllmain.cpp
```diff
@@ -102,6 +102,8 @@ void Shutdown()
 }
 
 BOOL APIENTRY DllMain(HMODULE mod, DWORD ul_reason_for_call, LPVOID) {
+	DisableThreadLibraryCalls(mod);
```
yamashi, line 105: indentation looks wrong

yamashi (changes requested):

src/overlay/Overlay_D3D12.cpp
WSSDude420: Not 100% sure now tbh. I'll double-check just to be sure after lineskips and such.

But tested all the scenarios I could think of (mixing different overlays running at once with this, only some,...)
Seems like there is no issue with how it is currently in my branch.

src/overlay/Overlay_D3D12.cpp
WSSDude420: Don't know honestly if I would make these regular functions. They have meaning only in context of this function, both.

But if you think it would be better, no problem to move both outside. But I really think it is unnecessary.
I also tried to made this lambdas in function format so no functors are in play.

src/dllmain.cpp
WSSDude420: no, is right here... problem is that { should be on next line, following your style. 

corrected.

src/overlay/Overlay_Hooks.cpp
WSSDude420: corrected

src/overlay/Overlay_D3D12.cpp
WSSDude420: corrected

src/overlay/Overlay_D3D12.cpp
WSSDude420: corrected

src/overlay/Overlay_D3D12.cpp
WSSDude420: corrected

src/overlay/Overlay_D3D12.cpp
WSSDude420: for now, corrected line skip here...

gonna leave it as lambda for now till you respond

src/overlay/Overlay_D3D12.cpp
WSSDude420: for now, corrected line skip here...

gonna leave it as lambda for now till you respond

WSSDude420:
Ok, waitting for response regarding those lambdas...

Gonna look up that Command Allocator usage in the mean time.

src/overlay/Overlay_D3D12.cpp
WSSDude420: You were correct, there could be an issue if two threads tried to use it at once, which could technically happen.

So fixed it.

Though it also seems like we should probably have separate command lists for each buffer as well...
Looking at how Reshade does it though, seems like we should be good in command lists regard with single list as we have now.

WSSDude420:
Well, hope it is better now...

Apart from lambdas, all was corrected now.
I also modified a bit log messages in Overlay, particularly ones I added in Overlay::InitializeD3D12(), as I found printing of swap chain argument a bit stupid on second glance.

src/overlay/Overlay_D3D12.cpp
```diff
@@ -29,82 +28,186 @@ BOOL CALLBACK EnumWindowsProcMy(HWND hwnd, LPARAM lParam)
...
+
+    for (auto& context : m_frameContexts)
+        m_pd3d12Device->CreateRenderTargetView(context.BackBuffer, nullptr, context.MainRenderTargetDescriptor);
+
+    IMGUI_CHECKVERSION();
+    ImGui::CreateContext();
+    ImGuiIO& io = ImGui::GetIO();
+    ImGui::StyleColorsDark();
+    io.Fonts->AddFontDefault();
+    io.Fonts->GetTexDataAsRGBA32(&pixels, &width, &height);
```
yamashi, line 171: I don't really see the user of this get?

yamashi (approved):

src/overlay/Overlay_D3D12.cpp
WSSDude420: Forgotten leftover from previous code I suppose. I believe I just copied this chunk in particular from former implementation. Will remove it ASAP.

WSSDude420:
Corrected previous issue with latest commit.

PR #308, opened 2020-12-26
Remove memory pool patch

WSSDude420:
I propose removal of memory pool patch, as it was apparently not useful at all.

See [this](https://www.cyberpunk.net/en/news/37166/hotfix-1-05) for reference.

> Removed the memory_pool_budgets.csv file. which was not connected with the final version of the game and had no influence on it (it was a leftover file used during the development to estimate memory usage. It had no effect on how much memory was actually allocated). Perceived performance increase after editing the file may have been related to restarting the game.


yamashi:
I removed it on my local branch already and did a big refactor so I am closing the PR because it will create needless conflicts.

PR #309, opened 2020-12-26
Uninitialize minhook and Overlay (if console true) during Shutdown()

WSSDude420:
Properly dispose of minhook on shutdown.


WSSDude420:
Also calling Overlay::Shutdown() now during Shutdown() now if console enabled.

KennnyTroll:
Hi have you seen my deposit? i made the same ... i think ... https://github.com/yamashi/CyberEngineTweaks/pull/297

PR #318, opened 2020-12-28
Add Windows 7 support to overlay

Mattiwatti:
This PR adds support for Windows 7 to the overlay, using Direct3D12 with D3D12On7 as discussed in #134.

Notes on implementation:
- I've mostly kept to the naming scheme Microsoft uses when it discusses D3D12 on Windows 7. This means that variables and functions specific to Windows 7 are suffixed with 'downlevel' to distinguish them from common D3D12 functionality.
- Two hooks have been added: `PresentD3D12Downlevel` and `CreateCommittedResource`. These functions are only hooked on Windows 7. The `PresentD3D12Downlevel` hook serves the same purpose as the current `Present` hook but for Windows 7, where Present is called on the device and not on the swap chain. The  `CreateCommittedResource` hook is required to find the backbuffer addresses, since there is no swap chain to query these from. This function unhooks automatically after D3D12 is initialized and it has served its purpose.
- `InitializeD3D12Downlevel` is the main function responsible for initializing D3D12 on Windows 7. It is largely based on the existing `InitializeD3D12`, but I felt there were enough differences to warrant splitting the two. I moved ImGui initialization to a shared function, since this works with no changes on Windows 7 and 10.
- Because there is no swap chain on D3D12 on Windows 7, it is not possible to obtain the current back buffer index in `Overlay::Render()`. This could potentially mean large amounts of flickering due to the overlay only being in 1/3rd of frames. I've worked around this by adding a simple counter that wraps to 0 after every 3 calls to `Present()`. This actuallly works surprisingly well, but note that I haven't tested this thorougly over multiple hours of play, and only on one machine. If the counter does turn out to de-sync over time, an alternative might be to reverse the game to obtain the current backbuffer index that way.

Problems/limitations:
- Changing the resolution or resizing the window will break the overlay on Windows 7, even if you restore the old resolution afterwards. This is due to the difficulty of finding the Present backbuffers when there is no swap chain to query them from. This is also the reason for the existence of the `CreateCommittedResource` hook. Theoretically this hook could be made to collect potential backbuffer addresses indefinitely after the game has started, but that still leaves the issue of figuring out what the correct moment to reinitialize D3D12 is after a resize has occurred. Besides this, Cyberpunk 2077 itself handles window resizes very poorly to begin with (try it on Windows 10 if you want), so I don't currently see much use for this.

### DIFF
--- a/src/kiero/kiero.cpp
+++ b/src/kiero/kiero.cpp
@@ -21,6 +21,7 @@
 #if KIERO_INCLUDE_D3D12
 # include <dxgi.h>
 # include <d3d12.h>
+# include "common/D3D12Downlevel.h"
 #endif
 
 #if KIERO_INCLUDE_OPENGL
@@ -46,6 +47,7 @@
 static kiero::RenderType::Enum g_renderType = kiero::RenderType::None;
 static uint150_t* g_methodsTable = NULL;
 static uintptr_t g_commandQueueOffset = 0;
+static bool g_isDownLevelDevice = false;
 
 kiero::Status::Enum kiero::init(RenderType::Enum _renderType)
 {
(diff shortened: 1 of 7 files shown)


WSSDude420:
Seems good to me. As i dont have review access, I'll share my thoughts here through comment.

- reset lambda could be now remade to something like InitializeD3D12Reset() or something and shared for both, so we get rid of some duplicite code
- we should probably check, as I do in InitializeD3D12, after initialization, if parameters match. There is always a potential of multiple swap chains and multiple command queues. We should be able to detect this, for debugging purposes + we probably dont want to draw imgui for command queue we are not bound to.

Dont know about other duplicite things... Those initialize functions look very similar I'd say, but whatever, probably is OK if owner does not oblige.

WSSDude420:
Actually, one more note from me - check indentation!

Seems your indentation settings are different, many lines are wrongly indented.

WSSDude420:
Ok, last few notes I hope...

- I would move this out of Render, as if it is truly like you say and we want to stay safe by incrementing it each present call... And want to do this "properly", so just for cmd queue we are bound to as said and should be added after initialization is completed as mentioned... As this would prevent Render from being called each present call potentially. https://github.com/yamashi/CyberEngineTweaks/pull/318/commits/d21b5f94cf2e7058de468b2a46e8579bb8a188e1#diff-e7620ecdaf8d13707a5172f4930ef637413ce791d90ac88918ca1bd7ff565dbaR340
probably would make it into some separate little function, or just simple lambda, and put it before init, as this has no bound to D3D12 it seems.
- I would personally probably pass that HWND to Initialize for downlevel too, and check it just to be sure it is indeed our target window for debugging purposes, like we check output window inside swapchain in regular Initialize. Didnt see it at first, but would be nice to do this I believe. Again, at least we know some more information if some problem happens. Should not be fatal though, so just a warning to the log so we know this happened, but may be not critical.

I know this all may kind of be a lot small BS one might think... But as we are not in full control of these, we should have as many safeguards in place as possible. D3D12 is kind of unpredictible, it being more LL API than HL, giving application more control over how and what is called.

src/kiero/kiero.cpp
```diff
@@ -536,8 +562,17 @@ kiero::Status::Enum kiero::init(RenderType::Enum _renderType)
 				commandList->Release();
 				commandList = NULL;
 
-				swapChain->Release();
-				swapChain = NULL;
+				if (swapChain != NULL)
```
yamashi, line 565: Could just use a ComPtr for direct3d types

yamashi (changes requested):

Mattiwatti:
> we should probably check, as I do in InitializeD3D12, after initialization, if parameters match. There is always a potential of multiple swap chains and multiple command queues. We should be able to detect this, for debugging purposes + we probably dont want to draw imgui for command queue we are not bound to.

I actually tried to do as much of this as possible, based on the existing `InitializeD3D12()`. However there are some things that I simply can't do: part of the `checkCmdQueue` lambda for example does not make sense on Windows 7 since there is no swap chain to use for obtaining a command queue. On the other hand I *do* have access to the command queue from the `PresentD3D12Downlevel` hook, since the `ID3D12CommandQueueDownlevel` passed as first argument can be converted to a regular `ID3D12CommandQueue`. Verifying that a command queue is actually the one we want is definitely something that should be done though, I think we are in agreement about that so I will add a check for this. The same thing again goes for validating the swap chain - inside the D3D12On7 initialization, there simply is no swap chain at all so this is a no-go. Same thing again for comparing the `HWND` of the swap chain... I think you get the idea.

This is why I said that, while I'm aware that the initialization functions are broadly similar, the lack of a swap chain in practice would make it so that such a 'unified' function would end up as a giant series of `if`/`else` statements just to deal with the quirks of D3D12 on Windows 7, especially w.r.t (1) the aforementioned absence of a swap chain, and (2) the convoluted hoops one needs to step through to find the backbuffers . But don't get me wrong; I don't actually have a very strong opinion regarding this, so if unifying the initialization functions is your preference then that's fine by me too.

> reset lambda could be now remade to something like InitializeD3D12Reset() or something and shared for both, so we get rid of some duplicite code

Agreed. I was thinking a plain member function would work fine here, and which could then be shared between all initialization functions (D3D12, D3D12On7, ImGui).

> I would move this out of Render, as if it is truly like you say and we want to stay safe by incrementing it each present call... And want to do this "properly", so just call Render for cmd queue we are bound to as said... As this check should be added into Initialize "after initialize" part (if m_initialized true).

Also agreed. `PresentD3D12Downlevel` would be a more suitable place to put this, in order to separate it from the rendering code. The only minor issue I see is that `Overlay::Render()` will still need to be told the correct buffer index to use on every present, via some function parameter or flag. But it is still an improvement I'd say.

> I would personally probably pass that HWND to Initialize for downlevel too, and check it just to be sure it is indeed our target window for debugging purposes, like we check output window inside swapchain in regular Initialize.

Sure. I will add this check.

src/kiero/kiero.cpp
Mattiwatti: Sure. But I saw that all the many other `IUnknown` derived types in kiero are also bare pointers, and I just went along with that in order not to disturb the peace of the file :wink:.

I can of course change this to a ComPtr, but in that case I'd prefer to make the change consistent across the entire file, and then make a new PR for that change. Thoughts?

WSSDude420:
We could make downbuffer index member variable if passing it around seems cumbersome. 

Statics/gloabls, sure not... But this would be one solution not requiring passing it through all the time which is still kind of proper.

PR #323, opened 2020-12-29
Use ImGuiListClipper to draw just visible output lines

### LINKED ISSUE #321: [Issue] Debug data dumping into Command Console Output
I don't know if anyone has posted this already or if anyone is aware, but when Debug is enabled, it will dump a bunch of the debug data into the output of the command console and it just keeps adding. This inturn adds load to the computer, which causes low fps and game crashes.

I tested seeing if clearing the output would relieve some of the stress. When I did (when I was at like 10 fps with the command console open), my fps spiked back up to normal (60-80fps), if not better levels. 

Not sure if you have plans on this or not, but a fix for this could prob be to add a limit (or a smaller limit, if one already exists) of how much of the command console output is stored. 

For now, I recommend to those who don't need it, to keeping debug disabled. And if you do enable debug, to only use it for a short session.

Edit:  I believe the other part of the problem is data from in game interactions being dumped into the Command Console output. I am not sure if there is a limit on this right now. If there was a way to disable all interaction data from showing up in the Command Console (only show data that was typed into the console), that may solve the biggest reason for crashes and lag.

WSSDude420:
As title says.

Fixes #321

### DIFF
--- a/src/overlay/Overlay.cpp
+++ b/src/overlay/Overlay.cpp
@@ -69,14 +69,20 @@ void Overlay::DrawImgui(IDXGISwapChain3* apSwapChain)
             ImVec2 listboxSize = ImGui::GetContentRegionAvail();
             listboxSize.y -= ImGui::GetFrameHeightWithSpacing();
             const auto result = ImGui::ListBoxHeader("", listboxSize);
-            for (auto& item : m_outputLines)
-                if (ImGui::Selectable(item.c_str()))
+            ImGuiListClipper clipper;
+            clipper.Begin(m_outputLines.size());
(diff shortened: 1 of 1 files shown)


WSSDude420:
Dont know why the check died?

Tried to rebuild release, as I just build debug always (no rebuilds).
Still builds...

Just a note, as I dont know what is the exact cause of the fail here... Seems it cannot find some mimalloc exports for whatever reason.

WSSDude420:
Lol, just forced to rerun check with force push and it worked... Something seems bugged with the check 🤔

PR #331, opened 2020-12-29
Additional logs, correct way we check for d3d12on7 according to MS docs

WSSDude420:


src/overlay/imgui_impl_dx12.cpp
```diff
@@ -454,30 +454,28 @@ bool    ImGui_ImplDX12_CreateDeviceObjects()
...
-                if (d3d12_dll == NULL)
-                    return false;
+            for (unsigned int i = 0; i < _countof(localD3d12Paths); i++)
+            {
+                d3d12_dll = LoadLibraryA(localD3d12Paths[i]);
+                if (d3d12_dll != NULL)
+                    break;
             }
+
+            // If failed, we should be on Windows 10. Attempt to load d3d12.dll from System32 only.
```
Mattiwatti, line 473: Nitpick: this comment is now wrong, since the `LOAD_LIBRARY_SEARCH_SYSTEM32` flag has been removed. It should just say we're on Windows 10.

WSSDude420: Fixed it for you before sleep :P

WSSDude420:
DONT MERGE YET!

Found some retarded comments now... Will correct them in a minute!

PR #338, opened 2020-12-30
D3D12 - detect resolution changes

WSSDude420:
For now, just a draft so @Mattiwatti knows what changes I'm making in real time better...

This pull request will try to resolve issues with resolution changes, which should help fix issues with fullscreen.

Note that this would be for D3D12, not D3D12on7. Latter will be worked on hopefully by Matti here.

### DIFF
--- a/src/overlay/Overlay.cpp
+++ b/src/overlay/Overlay.cpp
@@ -108,7 +108,11 @@ void Overlay::DrawImgui()
         }
 
         ImGui::SetNextItemWidth(ImGui::GetContentRegionAvail().x);
-        ImGui::SetKeyboardFocusHere();
+        if (Get().m_toggled)
+        {
+            ImGui::SetKeyboardFocusHere();
+            Get().m_toggled = false;
+        }
         const auto execute = ImGui::InputText("", command, std::size(command), ImGuiInputTextFlags_EnterReturnsTrue);
         ImGui::SetItemDefaultFocus();
         if (execute)
@@ -371,8 +375,11 @@ void Overlay::Toggle()
 
     while(true)
     {
-        if (m_enabled && ShowCursor(TRUE) >= 0)
+        if (m_enabled && ShowCursor(TRUE) >= 0) 
+        {
+            m_toggled = true;
             break;
+        }
         if (!m_enabled && ShowCursor(FALSE) < 0)
             break;
     }
--- a/src/overlay/Overlay.cpp
+++ b/src/overlay/Overlay.cpp
@@ -135,6 +135,15 @@ LRESULT APIENTRY Overlay::WndProc(HWND hwnd, UINT uMsg, WPARAM wParam, LPARAM lP
         return 0;
     }
 
+    if (uMsg == WM_WINDOWPOSCHANGED) 
+    {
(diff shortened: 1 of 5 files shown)


chendo:
This appears to be occasionally causing crashes for me whenever CP2077 loses focus. It either crashes at https://github.com/yamashi/CyberEngineTweaks/blob/master/src/overlay/imgui_impl_dx12.cpp#L350 or it crashes with bad memory access.

Dual monitor setup on Windows 10 Home.

WSSDude420:
Funny, tried to crash it religiously yesterday and could not...

Can you make it crash consistently with some action at specific time?
Would be easier to reproduce. I was alt+tabbing and clicking in browser in another window with both, console open and closed, during stsrtup, menu, settings, game,... Clicking back into window, etc.
Just losing and regaining focus as you said and no crashes for me in any scenario here 🤔

WSSDude420:
Anyway, I see that it tries to reset too many times before it can reinitialize, I believe this is the main cause of the issue here... I believe it resets state during it trying to initialize which leads to crash (seen it happen)
Question is how exactly did you manage to trigger multiple resets at once lol...

chendo:
I thought it was my code for while until it gave me a crash dialog that pointed to that line heh. It seems easier to cause this to happen during startup. I know I was able to crash it when hitting Windows key, as well as Alt Tab etc but it wasn't really consistent. I do have my secondary monitor in portrait though, not sure how relevant it is.

I also was mostly running in Borderless Fullscreen. Going Fullscreen seems to helped but I still ran into crashes at some point. I'll try to get back to you with repro steps on my machine when I get the chance, otherwise I can test probable fixes

WSSDude420:
Ah, ok, found one way to consistently crash it on my machine now...

During startup in any windowed mode, there is few seconds that mouse is visible if game has not yet lost focus.

It seems to crash while mouse is visible during said startup. At least for me...

As soon as the mouse disappears though, it seems to not crash anymore for me again.

May be some correlation there.

WSSDude420:
Also, problem seems to be during startup this particular event - having focus, mouse is visible, losing focus while in this state.

If it did not have any focus during startup to begin with, this seems to not materialize for me.

WSSDude420:
At least something to go by...

Well, anyway... I have something better in mind hopefully. Problem is that this was meant as kind of a workaround, as this should fire up on both, Windows 7 and Windows 10. 

My other change I had in mind now that I'm working on should be the most reliable way on Windows 10 to catch these, but we than miss out on Windows 7...

As from yesterday tests there, it seems reset event @Mattiwatti put there for Windows 7 users is not firing, only mine.

WSSDude420:
Well, I have super-reliable hook now for Windows 10 prepared.
At least in regards to resize events.

Question is, what with Windows 7? yamashi frankly does not care much from what I got from him in this regards...
I don't blame him tbh, it is quite a small percentage of our users... + hooking there in general needs to be performed in other way.

WSSDude420:
@chendo could you maybe test again with changes I currently have in draft here? #345

Feel free to pull it for testing, should be no conflicts there with master.

chendo:
@WSSDude420 I haven't managed to crash it so far! 👍

PR #344, opened 2020-12-31
Allow calling functions that don't expect ScriptGameInstance

chendo:
I've added some logic that checks if the first argument of `pFunc` expects a `ScriptGameInstance` before injecting the instance.

This enables calling of functions that don't require a `ScriptGameInstance`.

For example, calling `Game.GetFunFact()` without this PR will return `Function 'Game.GetFunFact' expects 4294967295 parameters, not 0`, whereas with the PR, it will call and return that function.

I haven't touched C++ for almost 20 years so apologies for any mistakes. Tested against master and 1.06.

Note: Build will fail because lua.org's SSL certificate expired earlier today.


WSSDude420:
I think I see nothing else wrong with this. If yamashi does not have other complaints, I believe this could be merged after this is fixed.

yamashi (approved):

PR #350, opened 2020-12-31
[WIP] Support properties in a backward compatible way

osulli:


osulli (commented):
L519 onwards will also need modifying to handle returrn types when running the LUA

PR #355, opened 2021-01-01
added math functions and preset teleport function

WakuriHisaki:
Added additional math functions that I couldnt seem to get access to using lua inside cp2077. These were required for creating an Euler from an Quaternion which can be used to set player rotation which I needed for the teleport script.

Also added a preset teleportForward(int distance) script that other people might find useful. Sorry if this is outside the scope of the project, but it seemed this project would be better to contain some general basic scripts than creating a seperate project.

Feedback is welcome, and sorry again if this is the wrong place for this type of contribution.

### DIFF
--- a/scripts/autoexec.lua
+++ b/scripts/autoexec.lua
@@ -7,3 +7,6 @@ Game['PlayFinisherSingle'] = Game['PlayFinisherSingle;GameInstance']
 Game['PPS'] = Game['PPS;GameInstance']
 
 print("Cyber Engine Tweaks autoexec.lua execution complete.")
+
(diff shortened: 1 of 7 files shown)


yamashi:
Hello,

Appreciate the PR but this is a mod in itself, I suggest you release this as a standalone, it's out of the scope of the project, we just expose tools for people like you, not end user tools.

Once released please send a link so I can add it to the list of mods ;)

WakuriHisaki:
Thank you for the clarification, I thought as much. I'll take your suggestion and start work on a seperate project for this.

If there is any testing or a specific focus point you'd like to explore then please let me know, i'd like to help and contribute where possible.

PR #359, opened 2021-01-02
More type support, more dumps, more accessible functions

expired6978:
Added Enum support with number<->Enum<->string
Added ref<->wref simple implicit cast
Added more dump support
Temp fix to allow searching function overloads


yamashi (approved):

yamashi:
Thanks for the PR!

PR #402, opened 2021-01-14
Make the console and imgui support utf-8 (Chinese)

Rorschach1219:
I hope you can release a Chinese version of the console. For the time being, I have made some changes to the code, and I will optimize the content of my changes if I have time.

I set the save format of Console.cpp to Unicode-codepage 1200
图片
202925zriygjupgrs7nhah


WSSDude420:
Do you have proper font management in place?
Do you have proper language management in place?

I dont think you have any single of these things in there judging by what you are committing here...

We cannot make it support one language but not others. We cannot load all glyphs for all languages at once for a font either with system ImGui is using. These need to be managed completely and properly. Without proper managment, you replace one problem with another... That is one giant thing for me here.

Another one, we want to have support for multiple fonts as this was much more requested than this right here. Already have something in place for this.
Now tell me how this what you have here would work with that. That is even more work to get done I'd say :) 

Personally, this is a big nope from me, but I'm not the owner of this repo so, we'll see how he responds.

WSSDude420:
Like I dont try now to completely disrespect this btw, just pointing out that what you did will just not work for general use case at all.

More of what my issue with this as a whole is.

If you can make some proper font/language managment that works well with shitty ImGui one which I dont think we can get around easily without even more work, nice.

But just replacing English with Chinnese and calling it Chinnese support is not right.

Rorschach1219:
> 
> 
> Like I dont try now to completely disrespect this btw, just pointing out that what you did will just not work for general use case at all.
> 
> More of what my issue with this as a whole is.
> 
> If you can make some proper font/language managment that works well with shitty ImGui one which I dont think we can get around easily without even more work, nice.
> 
> But just replacing English with Chinnese and calling it Chinnese support is not right.

I agree with what you said, I will close this request.
I know my behavior is too low-level, and I can understand your thoughts and opinions

Issue #399, opened 2021-01-11
[Bug] ImGui.PushStyleVar() ImGui.PopStyleVar() doesn't work


Nats-ji:
```lua
registerForEvent("onDraw", function()
    if (ImGui.Begin("Test Window")) then
        ImGui.PushStyleVar(ImGuiStyleVar.Alpha, 0.5)
        ImGui.PushStyleVar(ImGuiStyleVar.WindowPadding, 0.2, 0.1)
        ImGui.Text("Test 1234567")
        ImGui.Text("Test 1234567")
        ImGui.Text("Test 1234567")
        ImGui.Text("Test 1234567")
        ImGui.Text("Test 1234567")
        ImGui.Text("Test 1234567")
        ImGui.Text("Test 1234567")
        ImGui.PopStyleVar(2)
    end
    ImGui.End()
end)
```
Code inside `ImGui.PushStyleVar()` and `ImGui.PopStyleVar()` won't render.

PR #403, opened 2021-01-14
Fixed PushStyleVar() and PopStyleVar() binding

Nats-ji:
Fixed `PushStyleVar()` and `PopStyleVar()` binding in `sol_imgui.h`

#399

### DIFF
--- a/src/sol_imgui/sol_imgui.h
+++ b/src/sol_imgui/sol_imgui.h
@@ -15,7 +15,7 @@ namespace sol_ImGui
             ImGui::End();
             return std::make_tuple(false, false);
         }
-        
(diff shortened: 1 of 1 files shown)


WSSDude420:
Real change starts at line 2129, others were just corrected spaces from what I looked at.

Well, looks good to me, considering styling of that header.... But kind of worries me what is there but not registered now, cause this fix is kind of stupid in that nature...

Writting note to go through sol_ImGui.h when I'm done with other stuff cause there are some constant issues with ImGui impl, many stupid ones like this...

yamashi:
Thanks for the pr!

PR #408, opened 2021-01-18
Added and fixed things

expired6978:
Added vtable dump (DEBUG)
Added reflection dump (DEBUG)
Fixed insufficient args on function calls
Fixed out params on global functions
Added NewObject call
Added DumpAllTypeNames call Paired with NewObject to create objects in handles e.g. NewObject('Vendor') vs NewObject('handle:Vendor')
Exposed more ItemID and TweakDBID properties
Fixed Handle usage


src/reverse/WeakReference.cpp
```diff
@@ -3,17 +3,23 @@
...
+    if (ref)
+    {
+        m_pType = ref->GetType();
+    }
 }
 
 RED4ext::ScriptInstance WeakReference::GetHandle()
 {
-    return m_weakHandle.instance;
+    auto ref = m_weakHandle.Lock();
```
yamashi, line 18: Wouldn't that mean we can still have use after free?

yamashi (approved):

PR #436, opened 2021-01-24
Expose spdlog.info to mods

WSSDude420:
Fix small oversight on my part :P


WSSDude420:
Actually, this would be moved in sandboxing to somewhere else, just realised...

So can be closed, redundant PR in the end.

PR #454, opened 2021-01-31
Proper sandboxing

WSSDude420:
What is included:
- each mod is (right now) assigned sqlite3 database implicitly, accessible through `db` lua object (will be changed later to on-demand through info in metadata file)
- introduced proper sandbox with LuaSandbox helper class
- bytecode cannot be loaded now (native modules included)
- require cache is reset on Reload all mods now
- require does not support '.' instead of '/' now (only breaking change, functions the same otherwise)
- all pathing is now relative to mod (includes all io functions, dofile and such included)
- stripped ImGui, json, io and dofile and such from console sandbox (only needed things left for it in)
- added dir command for mods to list directory contents (requested by @Nats-ji )
- errors that happen in Game, game class methods,... should now log into proper environment (to console when executed from console, to mod log when executed from mod)


src/Utils.h
```diff
@@ -1,7 +1,32 @@
...
 void trim(std::string& s);
 
 spdlog::sink_ptr CreateCustomSinkST(std::function<void(const std::string&)> aSinkItHandler, std::function<void()> aFlushHandler = nullptr);
 spdlog::sink_ptr CreateCustomSinkMT(std::function<void(const std::string&)> aSinkItHandler, std::function<void()> aFlushHandler = nullptr);
-std::shared_ptr<spdlog::logger> CreateLogger(const std::filesystem::path& aPath, const std::string& aID, spdlog::sink_ptr aExtraSink = nullptr, const std::string& aPattern = "[%Y-%m-%d %H:%M:%S UTC%z] [%l] %v");
\ No newline at end of file
+std::shared_ptr<spdlog::logger> CreateLogger(const std::filesystem::path& aPath, const std::string& aID, spdlog::sink_ptr aExtraSink = nullptr, const std::string& aPattern = "[%Y-%m-%d %H:%M:%S UTC%z] [%l] %v");
+
+template <typename ...Args>
+void TryLuaFunction(std::shared_ptr<spdlog::logger> logger, sol::function func, Args... args)
```
yamashi, line 12: This shouldn't be in Utils I think

src/scripting/LuaSandbox.cpp
```diff
@@ -0,0 +1,455 @@
...
+}
+
+void LuaSandbox::Initialize(sol::state_view aStateView)
+{
+    // initialize state + environment first
+    m_lua = aStateView;
+    m_env = { aStateView, sol::create };
+
+    // copy whitelisted things from global table
+    constexpr std::array<std::string_view, 55> whitelistedGlobals =
```
yamashi, line 19: I think this is a bug waiting to happen, I think it would be best to use a standard C array, std::size will give us the number of elements and we don't have to give it a size manually.

src/scripting/LuaSandbox.cpp
```diff
@@ -0,0 +1,455 @@
...
+    assert(id < m_sandboxes.size());
+    return m_sandboxes[id];
+}
+
+void LuaSandbox::InitializeExtraLibsForSandbox(Sandbox& aSandbox) const
+{
+    auto& sbEnv = aSandbox.GetEnvironment();
+
+    // copy extra whitelisted libs from global table
+    constexpr std::array<std::string_view, 23> whitelistedTables =
```
yamashi, line 199: Standard C array

yamashi (changes requested):

src/Utils.h
WSSDude420: Moved into ScriptContext.cpp, as that was only place it was used in the end (wanted to use it elsewhere too before but found out that is unusable in other cases....)
Hope it is in better place now :P

src/scripting/LuaSandbox.cpp
WSSDude420: std::size is actually not needed as it looks like static arrays are supported by ranged-loops. Didnt know, wanted to use those (so there is not too many things to change when adding something to whitelist) which is the reason I went with std::arrays before. 
I guess this is truly better though, one less thing to think about and this now fully achieves what I wanted to do with this :P

PR #456, opened 2021-02-01
Fix HK handler causing sol state corruption leading to crash

WSSDude420:
Found a way to solve this elegantly enough I believe :P 

Is fully compatible with #454 and #453

### DIFF
--- a/src/CET.cpp
+++ b/src/CET.cpp
@@ -56,16 +56,18 @@ CET::CET()
     : m_options(m_paths)
     , m_bindings(m_paths)
     , m_window(&m_overlay, &m_bindings, &m_d3d12)
-    , m_d3d12(m_window, m_paths, m_options, m_bindings)
+    , m_d3d12(m_window, m_paths, m_options)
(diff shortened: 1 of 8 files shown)


src/CET.cpp
```diff
@@ -56,7 +56,7 @@ CET::CET()
     : m_options(m_paths)
     , m_bindings(m_paths)
     , m_window(&m_overlay, &m_bindings, &m_d3d12)
-    , m_d3d12(m_window, m_paths, m_options)
+    , m_d3d12(m_window, m_paths, m_options, m_bindings)
```
yamashi, line 59: Instead of doing this, you could just register a function in m_d3d12.OnUpdate ?

yamashi (approved):

src/CET.cpp
WSSDude420: hope that is better now :P 
didnt know where to put it before... but i think this would be the most proper place for it.
dont know if you like my approach though. found it better than having this connect performed in CET master

yamashi (approved):

PR #462, opened 2021-02-03
Converter::ToRED - LuaRED implicit conversion should only be for numbers

Sombra0xCC:
This fixes the annoying "expecting userdata got XX"
it also lets Scripting::ToRED return nullptr in stackType.value so you could add logs in your code for the expected type.
It shouldn't cause any issues with any mod

I do it in TweakDB::SetFlat
```cpp
RED4ext::CStackType stackType = Scripting::ToRED(aValue, stackTypeCurrent.type, &s_scratchMemory);
if (stackType.value == nullptr)
{
    RED4ext::CName typeName;
    stackTypeCurrent.type->GetName(typeName);
    spdlog::get("scripting")->info("[TweakDB::SetFlat] Failed to convert value. Expecting: {}", typeName.ToString());
    return false;
}
```

### DIFF
--- a/src/reverse/LuaRED.h
+++ b/src/reverse/LuaRED.h
@@ -2,7 +2,7 @@
 
 #include "common/Meta.h"
 
-template<class T, FixedString REDName>
+template<class T, FixedString REDName, bool CheckObjectType = !std::is_arithmetic_v<T>>
 struct LuaRED
 {
     static constexpr char const* Name = REDName;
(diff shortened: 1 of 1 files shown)


WSSDude420 (commented):
I dont understand honestly why LuaRedNumber was added just for an extra if. Is it really necessary to have this class? 

Kind of not getting the point of it tbh, seems redundant to make this distinction for numbers but maybe I'm missing something :)

It just looks redundant, maybe it has its meaning. But tbh, if this is only difference between the two, dunno if it would not be better still to have just LuaRED and have extra template param to make this extra check active or not.

yamashi:
Waiting for the issues @WSSDude420 mentioned to be addressed. Thanks for your work and contributions :)

Sombra0xCC:
I don't know why i did that..
Fixed PR.

WSSDude420 (approved):

Issue #465, opened 2021-02-06
Expose some keybind API to mods so they can read what is bound to some action (if anything)


mesakomarevich:
Looking through the code, I can't see any clear way to figure out what keys are bound to what actions. If this functionality does not exist, I think implementing it could be useful for any mod that implements a GUI with, and possibly other cases.

For instance [NPC Tool](https://www.nexusmods.com/cyberpunk2077/mods/898?tab=description) hard-codes the key-binds it uses in it's GUI. This mod is currently broken, but I've fixed it by replacing calls to `ImGui.IsKeyPressed()` with calls to `registerHotkey()` only to find out that I can't seem to update the GUI with the hotkeys chosen by the user.

enzo522:
[SimpleAppearance.zip](https://github.com/yamashi/CyberEngineTweaks/files/5949229/SimpleAppearance.zip)

Hello, mesakomarevich.
If you only want to NPC Tool to work again, try above mod which I've implemented.
I've made it because NPC Tool doesn't exist after update.
Otherwise, just retreive what you want then merge to your project.

The mod I attached has Hotkey functions.
You could set them in Cyber Engine Tweaks overlay.
I recommend you to set "Select Next Item Hotkey" as PageDown, "Select Prev Item Hotkey" as PageUp button.

mesakomarevich:
Hey, @enzo522, I'll take a look at your mod. I have however, already fixed NPC tool for myself and provided a fix on the mod page, but my main concern here is being able to figure out which key the user has bound to which action so that it can be displayed in the UI.

enzo522:
> Hey, @enzo522, I'll take a look at your mod. I have however, already fixed NPC tool for myself and provided a fix on the mod page, but my main concern here is being able to figure out which key the user has bound to which action so that it can be displayed in the UI.

Ah, ha. Great.
It would be better to refer to Cyber Engine Tweak implementation.
Check: https://github.com/yamashi/CyberEngineTweaks/blob/v1.10.2/src/VKBindings.cpp

As you can see "VKBindings::GetSpecialKeyName" function, I think there is no more elegant way to display string of pressed key.

mesakomarevich:
[...]

PR #512, opened 2021-03-08
Enhance VKBindings + modify everything else accordingly

### LINKED ISSUE #424: registerInput handler to allow for more regular inputs (KeyDown+KeyUp events)
_Amazing work on the new update. Love the new console popup with settings + ability to bind mod keys!_

Testing it out now with the following code:
```
registerHotkey('KEYPRESS_TEST','Test Key Press Event', function()
	print('Key press event fired!')
end)
```

The bind appears in the new console popup and I can bind a key.
If I press the key, I see the message successfully printed in the console, however if I HOLD the key, it only plays the event once.

Is there a way to have it repeat the event?
Otherwise, I think we need a function for "key down" and "key up" so that we can call some function loop over and over until the "key up" function terminates it.

WSSDude420:
Fix #424 and #465

Changes:
- scroll wheel support (in both axes)
- added registerInput for single inputs
- renamed Hotkeys to Bindings (added needed compatibility checks)
- as we have two input styles now, hotkeys are marked with [HK] prefix inside bindings menu
- add functions to get key bind string
- expose IsBound and GetBind (string version) to particular Sandbox environments

### DIFF
--- a/src/Options.cpp
+++ b/src/Options.cpp
@@ -33,6 +33,18 @@ void Options::Load()
             if (OverlayKeyBind == 0)
                 IsFirstLaunch = true; // is for sure in this case
 
+            if (exists(m_paths.CETRoot() / "hotkeys.json"))
+            {
+                // encoded key bind was 32-bit number in old config, convert it to new 64-bit format
+                OverlayKeyBind =
+                {
+                      ((OverlayKeyBind & 0x000000FF) << 8*0)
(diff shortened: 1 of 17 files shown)


WSSDude420:
DONT MERGE YET!

Forgot to mark it as draft

WSSDude420:
As merging this or #514 would break merge of the other, i premerged it and fixed conflicts.

Issue #473, opened 2021-02-11
New versions keeps crashing my game. Old not.


Edenia3:
Yea. Just figured it out.
There is a bug with table.insert and database in new versions, it causes crashes.

WSSDude420:
it seems to be linked to something else, we already know what... ill keep this open till it is fixed as a reminder of sorts.

issue lies within require, more specifically require within require it seems. dont know what is causing this, had not yet time to debug it all, but issue is known and also is the cause.

if this by a chance is not right (you can reproduce it somehow without multiple requires) let me know + share a snippet if you can cause that would be new issue in that case we are unaware of.

WSSDude420:
```With current CET it threw the error. Moving the content of labels.lua to the UI module made it work again. So I thought it must have something to do with how the labels module is loaded. Then I put the "require" outside the "UI.Create" function and suddenly everything is working again.```

[error_sample.zip](https://github.com/yamashi/CyberEngineTweaks/files/5982304/error_sample.zip)

Issue #494, opened 2021-02-15
Weird issue which cannot be explained without sample...


WSSDude420:
Reported by @psiberx some time ago and forgotten about... May be not a case anymore, need to test this and see what's up.

this single line `if ready then`, where `ready` is a boolean, causes this error in the log:
```
[2021-02-09 12:20:58 UTC+03:00] [error] stack index -1, expected table, received boolean: value cannot not have a valid environment (type check failed in constructor)
stack traceback:
    [C]: in metamethod 'index'
    ...\bin\x64\plugins\cyber_engine_tweaks\mods\bug01\init.lua:33: in field 'callback'
    ...7\bin\x64\plugins\cyber_engine_tweaks\mods\bug01\mod.lua:152: in field 'onUpdateEvent'
    ...\bin\x64\plugins\cyber_engine_tweaks\mods\bug01\init.lua:27: in function <...\bin\x64\plugins\cyber_engine_tweaks\mods\bug01\init.lua:11>
```
putting one extra line `print(type(ready))` right before `if ready then` eliminates the error
really hard to explain it, need to see the sample
just wait till the main menu, this is where `ready == true` and error starts popping in the log

[wtf02.zip](https://github.com/yamashi/CyberEngineTweaks/files/5982549/wtf02.zip)

WSSDude420:
Looks like this issue is still present

WSSDude420:
There should be this inside init.lua
```lua
mod = require('mod')
ready = false
```
instead of 
```lua
local mod = require('mod')
local player, statsSystem, equipmentSystem
local ready = false
```

PR #514, opened 2021-03-09
Fixes for sandboxed environments

### LINKED ISSUE #502: loadstring, loadfile, dofile and require (potentially others) won't work from registerForEvent/registerHotkey callbacks
test.lua:
```lua
print('hello')
```
init.lua:
```lua
registerForEvent('onUpdate', function(deltaTime)
  dofile('test.lua')
end)
```
result:
```
[2021-02-19 21:12:25 UTC+01:00] [error] stack index -1, expected table, received boolean: value cannot not have a valid environment (type check failed in constructor)
stack traceback:
    [C]: in global 'dofile'
    ...077\bin\x64\plugins\cyber_engine_tweaks\mods\df\init.lua:38: in function <...077\bin\x64\plugins\cyber_engine_tweaks\mods\df\init.lua:32>
```

dofile can be replaced by loadstring, loadfile, require also (with appropriate changes), leads to same outcome.

WSSDude420:
Fix dofile, require, loadfile, loadstring (#402)
Fix for recursive require calls (#473)
Fix issue with local vs. global vars (#494)
Disallow init.lua in binary form (was not checked at all before)

### DIFF
--- a/src/scripting/FunctionOverride.cpp
+++ b/src/scripting/FunctionOverride.cpp
@@ -48,7 +48,7 @@ bool FunctionOverride::HookRunPureScriptFunction(RED4ext::CClassFunction* apFunc
         const auto& calls = itor->second.Calls;
         for (const auto& call : calls)
         {
-            const auto result = call->ScriptFunction(as_args(args), call->Environment);
+            const auto result = call->ScriptFunction(as_args(args));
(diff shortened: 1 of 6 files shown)


WSSDude420:
Also fixes #473 from what I tested now

PR #523, opened 2021-03-27
Better isolation of sandboxed environments

WSSDude420:
This PR prevents users from overwriting tables of another mod by accident.

Deep copying is avoided inside LuaSandbox::Initialize on purpose, as that function just creates base templated environment for each Sandbox instance. Sandbox instance makes deep copy during construction.

Deep copy does not take into account duplicates, which should not really happen in our use case I believe. Did not found anything which would break cause of wrong copy.
We do not deep copy keys, so keys which are tables will not get properly copied (we do this only inside DeepCopySolObject function), but again, should not matter for our use case. Can be added relatively easily if required, have previous code saved which had this function in.

Before this, something like this could be done, affecting other mods in the process:
```lua
ImGui.Begin = function(name) spdlog.info("PWND!") end
```

Didnt test the snippet, may be wrong, but you get the picture...

Cleaned-up version of #522 

---------------------

Also adds Double inside meta visitor of Converter ( @psiberx )

### DIFF
--- a/src/Utils.cpp
+++ b/src/Utils.cpp
@@ -109,3 +109,25 @@ sol::object DeepCopySolObject(sol::object aObj, const sol::state_view& aStateVie
     copy[sol::metatable_key] = src[sol::metatable_key];
     return copy;
 }
+
+// makes sol object immutable when accessed from lua
+void MakeSolObjectImmutable(sol::object aObj, const sol::state_view& aStateView)
+{
+    if (aObj.get_type() != sol::type::table && aObj.get_type() != sol::type::userdata)
+        return;
+
+    sol::table target = aObj;
+    sol::table metatable = target.get_or(sol::metatable_key, sol::nil);
+
+    if (metatable == sol::nil)
+    {
+        metatable = {aStateView, sol::create};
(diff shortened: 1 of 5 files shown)


psiberx:
Alongside deep copy this PR adds extra protection for usertypes (`Vector4`, `Enum`, etc.) and userdatas (main `Game` instance) that are shared between mod environments. This prevents adding new properties, overriding existing properties, and manipulating metatables.

An example of what was possible before PR:
```lua
getmetatable(Game).__index = function() 
    print('pwnd') 
    return function() end
end

Game.GetPlayer() -- pwnd
Game.GetStatsSystem() -- pwnd
```

bonanov:
So, does this PR breaks the only way for mods to talk to each other?

WSSDude420:
@bonanov no?

GetMod is still there and functions in exactly the same way.

This was done to prevent accidentaly screwing environment for other mods. If you need to communicste with some mod, you should have used GetMod, which was and still is there.

bonanov:
@WSSDude420 oh, i was not aware of GetMod. Thank you!

PR #532, opened 2021-04-01
Fix LuaRED ToLua conversion function for 64-bit integral types

WSSDude420:
Hacky solution for LuaRED ToLua conversion of 64-bit integral types.

Abuses the fact that LuaJIT internally uses cdata for these types from scripts.

Thanks @psiberx for helping with this.


WSSDude420:
I suppose you have cancelled the run lol, was not me :D

Is same code as in the first run basically, so yeah... no need to run checks :P

PR #529, opened 2021-03-30
Big UI changes (full change list in description)

WSSDude420:
- Made Overlay key available only in VKBindings system to prevent duplication
- Moved Overlay key binding to Bindings widget
- First time launch now has modal popup where user needs to bind Overlay key
- Settings and Bindings widgets have now modal popups when there are unsaved changes
- Added option to disable removing of dead key bindings (good for debugging)
- Added toggle for ImGui Diagnostic Window widget
- Reworked Settings widget (has two tabs, one for patches and one for developer options)
- Reworked how widgets are enabled/disabled in Overlay
- Mods now do not draw on first launch until Overlay key is bound
- Created new helper widget for unapplied changes modal dialog
- Made BindWidget more general
- Modal dialogs will block LuaVM draw temporarily to prevent ImGui context breakage (beware that Update still runs! dont draw shit in Update please!)
- Added optional X offset argument to BindWidget
- Added option to enable ImGui assertions at will (default is on)
- Nicer formatting for headings of Bindings widget

Already merged from split PRs:
- Fixed issue with console history in Console widget (found by @Sombra0xCC )
- Fix single inputs in VKBindings
- Fix ImGui.TreePop unavailable in Lua
- Added optional X offset argument to BoolWidget
- Added DelayedSetTrapInputInImGui function to be used from within Overlay (needed because of design changes)
- More smaller internal things not listed here before but listed in respective PRs

PR #536, opened 2021-04-02
Smaller fixes and additions

WSSDude420:
- Added DelayedSetTrapInputInImGui function so we are able to trap input asynchronously at the right time (inside message queue seems to work always) [used by #529 ]
- Added ability to block mod drawing into LuaVM [used by #529]
- Fixed wrong TweakDB assertion
- Fixed DeepCopySolObject utility function for nil values
- Fixed issue with console history not copying over null terminator

### DIFF
--- a/src/d3d12/D3D12.h
+++ b/src/d3d12/D3D12.h
@@ -15,8 +15,9 @@ struct D3D12
 
     D3D12(Window& aWindow, Paths& aPaths, Options& aOptions);
     ~D3D12();
-    
+
     void SetTrapInputInImGui(bool aEnabled);
+    void DelayedSetTrapInputInImGui(bool aEnabled);
     [[nodiscard]] bool IsTrapInputInImGui() const noexcept { return m_trapInputInImGui; }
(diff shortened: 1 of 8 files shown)


src/d3d12/D3D12.h
```diff
@@ -77,4 +78,7 @@ struct D3D12
     Paths& m_paths;
     Window& m_window;
     Options& m_options;
+    
+    std::atomic_bool m_delayedTrapInput{ false };
```
yamashi, line 82: Those should be wrapped in setters/getters

yamashi (changes requested):

src/d3d12/D3D12.h
WSSDude420: I think I dont understand you now?
These are set from DelayedSetTrapInputImGui function which was added.

I dont think there should be ever need for gets for this. Setter should be in place.

WSSDude420: Can change it but like I dont understand now what would you like changed exactly :)

yamashi: Ah right it's because you access them from a static

PR #537, opened 2021-04-02
Minor additions to UI

WSSDude420:
- Changed some ImGui::Text calls to ImGui::TextUnformatted to not unnecessarily invoke snprintf
- Added Reset Changes button to Bindings widget [new function is also used by #529 ]
- Added new helper widget for unsaved changes modal [used by #529 ]
- Made existing widgets a bit better [rest of additions are coupled inside #529 as that would not compile on its own in this state]
- Change definition for ImGui Text Lua binding to point to TextUnformatted, as we do not support formatting in current bindings implementation anyway. Should avoid some overhead from snprintf, as Text is commonly used element.

### DIFF
--- a/src/overlay/widgets/HelperWidgets.cpp
+++ b/src/overlay/widgets/HelperWidgets.cpp
@@ -87,25 +87,25 @@ namespace HelperWidgets
         }
     }
 
-    bool BoolWidget(const std::string& label, bool& current, bool saved, float offset_x)
+    bool BoolWidget(const std::string& aLabel, bool& aCurrent, bool aSaved, float aOffsetX)
     {
         ImVec4 curTextColor = ImGui::GetStyleColorVec4(ImGuiCol_Text);
-        if (current != saved)
+        if (aCurrent != aSaved)
             curTextColor = ImVec4(1.0f, 1.0f, 0.0f, 1.0f);
 
         ImGui::AlignTextToFramePadding();
 
-        ImGui::SetCursorPosX(ImGui::GetCursorPosX() + offset_x);
+        ImGui::SetCursorPosX(ImGui::GetCursorPosX() + aOffsetX);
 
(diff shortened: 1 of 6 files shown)


src/overlay/widgets/HelperWidgets.cpp
```diff
@@ -84,19 +87,72 @@ namespace HelperWidgets
         }
     }
 
-    void BoolWidget(const std::string& label, bool& current, bool saved)
+    bool BoolWidget(const std::string& label, bool& current, bool saved, float offset_x)
```
yamashi, line 90: Naming conventions

src/overlay/widgets/HelperWidgets.h
```diff
@@ -6,5 +6,9 @@ namespace HelperWidgets
 {
 WidgetID ToolbarWidget();
 void BindWidget(VKBindInfo& aVKBindInfo, const std::string& acId);
-void BoolWidget(const std::string& label, bool& current, bool saved);
+bool BoolWidget(const std::string& label, bool& current, bool saved, float offset_x = 0.0f);
```
yamashi, line 9: Naming conventions

yamashi (changes requested):

src/overlay/widgets/HelperWidgets.cpp
WSSDude420: Changed 👍

src/overlay/widgets/HelperWidgets.h
WSSDude420: Changed 👍

WSSDude420:
Had to add this one in real quick into this...

This was not in #529, this is something that now came to my head.

Is unnecessary to have calls into ImGui::Text when you dont want to really format things. Which you currently cannot from mods.

yamashi (approved):

PR #538, opened 2021-04-02
Override ImGui asserts by custom definition for more granular control

WSSDude420:
This PR will reenable assertions in release. Control for these new assertions can be found in #529 .


WSSDude420:
Ok, this requires that TiltedCore update to v0.2.2
So this is alright, just package update is not included here :)

This + #536 should compile without issues.

WSSDude420:
Ok, it is also missing definition for that global...

WSSDude420:
In case you were asking why I did this with a global and not check inside assertion function...

I did it that way before. I then realised though that each expression would get evaluated no matter what...

Having global like this against which it is compared BEFORE doing rest of the things should skip those checks if assertions are disabled, avoiding unnecessary code execution.

yamashi (approved):

PR #541, opened 2021-04-04
Update README.md

WSSDude420:
Also, you could probably change description of repo to what we have on Nexus, it reflects better what this does and provides I believe :)

"Scripting framework for modders and quality of life fixes."

I mean this specifically...
Capture

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,25 +9,38 @@
 
 ## What's this?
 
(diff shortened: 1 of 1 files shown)


WSSDude420:
Canceled workflow cause is useless to let it run for readme update.

PR #543, opened 2021-04-05
Reorder roll, pitch, and yaw to be consistent with the internal order

Hawkpath:
EulerAngles::ToString had the roll and yaw values swapped because of incorrect parameter order in [Scripting.cpp](https://github.com/yamashi/CyberEngineTweaks/blob/master/src/scripting/Scripting.cpp#L183). I fixed this by rearranging the parameter order everywhere to be consistent with [EulerAngles.hpp](https://github.com/yamashi/RED4ext.SDK/blob/6438a96c0fe1e6e00a2815c97bbf51b1b0bdc9ee/include/RED4ext/Types/generated/EulerAngles.hpp#L16) (roll, pitch, yaw).

### DIFF
--- a/src/reverse/BasicTypes.cpp
+++ b/src/reverse/BasicTypes.cpp
@@ -17,7 +17,7 @@ std::string Vector4::ToString() const noexcept
 
 std::string EulerAngles::ToString() const noexcept
 {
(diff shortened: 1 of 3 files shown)


yamashi:
Thanks for your contribution :)

PR #544, opened 2021-04-08
Fixes for out params, mod database, and lua operators for basic types

psiberx:
1. Fixed out params handling when out params are not in the end of the param list.  
```
import function GetCurveValue(out x : Float, out y : Float, curveName : CName, isDebug : Bool);
import function GetObjectClosestToCrosshair(instigator : whandle:gameObject, out angleDistance : EulerAngles, query : gameTargetSearchQuery) : handle:gameObject;
```

2. Fixed CClassConverter for nils. This call `args[i] = Scripting::ToRED(sol::nil, apFunc->params[i]->type, &s_scratchMemory);` for creating a placeholder for the out param was resulting in a `nullptr` when param is not a handle.
```
import function LookAt(instigator : whandle:gameObject, out params : gameaimAssistAimRequest);
```

3. Comparison operators for basic types and concatenation for TweakDBID:  
```lua
print(Vector3.new(1, 2, 3) == ToVector3{ x = 1, y = 2, z = 3 }) -- true
print(Enum.new("gameGameVersion", "Current") == Enum.new("gameGameVersion", "CP77_Patch_1_2")) -- true
print(TweakDBID.new("Items.Preset_Igla_Default") + ".displayName" == TweakDBID.new("Items.Preset_Igla_Default.displayName")) -- true
```

4. Close mod DB on shutdown. If DB is not closed, after reloading the mods, the DB file may be locked for a while, so any attempts to change the DB in onInit will fail.

5.  Removed Vector3 duplicate (the usertype was defined twice in Scripting.cpp) and moved Enum definition.

6.  Fixed incorrect assert in VKBindings.cpp.

### DIFF
--- a/src/reverse/Converter.h
+++ b/src/reverse/Converter.h
@@ -182,7 +182,7 @@ struct ClassConverter : LuaRED<ClassReference, "ClassReference">
             else
             {
                 result.value = nullptr;
-			}
+            }
         }
         else
         {
--- a/src/reverse/Converter.h
+++ b/src/reverse/Converter.h
@@ -172,6 +172,18 @@ struct ClassConverter : LuaRED<ClassReference, "ClassReference">
         {
             result.value = aObject.as<ClassReference*>()->GetHandle();
         }
+        else if (aObject == sol::nil)
+        {
+            RED4ext::CClass* pClass = static_cast<RED4ext::CClass*>(apRtti);
+            if (pClass && !pClass->flags.isAbstract)
+            {
+                result.value = pClass->AllocInstance();
+            }
+            else
+            {
+                result.value = nullptr;
+			}
+        }
         else
         {
             result.value = nullptr;
--- a/src/VKBindings.cpp
+++ b/src/VKBindings.cpp
@@ -99,7 +99,7 @@ std::vector<VKBindInfo> VKBindings::InitializeMods(std::vector<VKBindInfo> aVKBi
     }
 
     // insert CET overlay bind info
-    assert(m_pOverlay); // overlay must be set before first use!
+    assert(m_cpOverlay); // overlay must be set before first use!
     const auto overlayKeyBind { m_cpOverlay->GetBind() };
     const auto overlayKeyCodeIt { m_idToBind.find(overlayKeyBind.ID) };
     const auto overlayKeyCode { (overlayKeyCodeIt == m_idToBind.cend()) ? (0) : (overlayKeyCodeIt->second) };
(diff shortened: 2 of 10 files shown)


src/reverse/Converter.h
```diff
@@ -172,6 +172,18 @@ struct ClassConverter : LuaRED<ClassReference, "ClassReference">
...
+        {
+            RED4ext::CClass* pClass = static_cast<RED4ext::CClass*>(apRtti);
+            if (pClass && !pClass->flags.isAbstract)
+            {
+                result.value = pClass->AllocInstance();
+            }
+            else
+            {
+                result.value = nullptr;
+			}
```
yamashi, line 185: weird

src/reverse/Converter.h
```diff
@@ -172,6 +172,18 @@ struct ClassConverter : LuaRED<ClassReference, "ClassReference">
         {
             result.value = aObject.as<ClassReference*>()->GetHandle();
         }
+        else if (aObject == sol::nil)
+        {
+            RED4ext::CClass* pClass = static_cast<RED4ext::CClass*>(apRtti);
+            if (pClass && !pClass->flags.isAbstract)
+            {
+                result.value = pClass->AllocInstance();
```
yamashi, line 180: I am not sure this is the right way to do this considering we don't release memory.
Also if I remember correctly AllocInstance just creates the memory but doesn't call the constructor

yamashi (changes requested):

src/reverse/Converter.h
psiberx: I copied the logic from `NewObject` for `pType->GetType() == RED4ext::ERTTIType::Class` case:
https://github.com/yamashi/CyberEngineTweaks/blob/master/src/scripting/Scripting.cpp#L284
https://github.com/yamashi/CyberEngineTweaks/blob/master/src/scripting/Scripting.cpp#L303
This should create an empty instance for out param, which will be added to the result set after execution `results.push_back(Scripting::ToLua(state, args[i]));` where `args[i]` is the result of conversion.

yamashi: Alright it's a bug that's more prevalent than I thought, will create another issue for this

src/reverse/Converter.h
psiberx: Incorrect indentation or logic?

yamashi: indentation

psiberx: Fixed

PR #548, opened 2021-04-17
Error logging for Observe / Override, implicit TweakDBID conversion, minor environment fixes

psiberx:
1. Added reporting for errors occurred in `Observe` / `Override` handlers.
This fails silently before fix:
```lua
Observe('gameObject', 'OnGameAttached', function()
    undefined.call()
end)
```

2. Added implicit string to `TweakDBID` conversion. 
```lua
print(Game.GetTransactionSystem():GetItemInSlot(Game.GetPlayer(), 'AttachmentSlots.WeaponRight'):GetItemID())
```

3. Removed redundant environment passing to Lua functions as an argument. This change addresses two issues:
a. Environment passed as an argument doesn't change the environment of the function, but goes straight to the called function as a table.
b. The function inherits environment when defined and the assignment is stored in the Lua state. There is no need to set environment for function before call unless it's intended to change the currently assigned environment to another one. Also it should be taken in account that any environment assignment is permanent and will not be reverted to previous environment in any implicit way.

### DIFF
--- a/src/reverse/Converter.cpp
+++ b/src/reverse/Converter.cpp
@@ -21,8 +21,8 @@ auto s_metaVisitor = [](auto... args) {
     LuaRED<Vector4, "Vector4">(),
     LuaRED<EulerAngles, "EulerAngles">(),
     LuaRED<ItemID, "gameItemID">(),
-    LuaRED<TweakDBID, "TweakDBID">(),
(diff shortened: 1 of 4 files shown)


yamashi:
Looks good! Thanks :)

PR #549, opened 2021-04-21
D3D12 - Ensure that ImGui_ImplDX12_CreateDeviceObjects never receives a nullptr command queue

PotatoOfDoom:
This small change should fix CET shutting down when executing D3D12::Initialize before D3D12::ExecuteCommandLists. This happens because ImGui relies on an existing commandQueue. Normally this isn't a problem because the game always calls those hooks in the correct order but some overlay software might call present before executecommandlist during initialization (In my case the specialK framework) and can cause cet to break. 
It even seems like the original code tries to take something like this into account but the order of the functions is mixed up.


WSSDude420 (approved):

yamashi:
Thanks for your contribution :)

PR #552, opened 2021-05-04
added TweakDB::SetFlats

Sombra0xCC:
TweakDB QoL
```lua
TweakDB:SetFlats('Character.Judy', {
  level = 99,
  tags = { 'Immortal' }
});
```

### DIFF
--- a/src/reverse/TweakDB.cpp
+++ b/src/reverse/TweakDB.cpp
@@ -208,6 +208,30 @@ sol::object TweakDB::GetFlat(TweakDBID aDBID)
     return Scripting::ToLua(state, stackType);
 }
 
+bool TweakDB::SetFlatsByName(const std::string& acRecordName, sol::table aTable, sol::this_environment aThisEnv)
+{
+    return SetFlats(TweakDBID(acRecordName), std::move(aTable), std::move(aThisEnv));
+}
+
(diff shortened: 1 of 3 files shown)


Sombra0xCC:
don't merge. aObject.is<uint64_t> is not working...

Sombra0xCC:
some problems with s64/u64:
1) can't check if the lua object is s64/u64, passing wrong data throws exception in strtoll
2) cant be casted to other int types
pic

other than that, everything is working.
Trying to convert an 'int' to 'string' now returns nullptr instead of throwing sol errors

psiberx:
It's a LuaJIT specifics. Lua 5.1 supports only doubles and no int64s. LuaJIT implements them as special cdata structure. It's so non-standard you can't even check for that type using sol. But it can be done using Lua API directly. 
The casting though might be done without checks with applying Lua's `tonumber()` first to get the double from cdata, and then casting the result.

psiberx:
I can merge this (eliminate CheckObjectType) with my changes to LuaRED to avoid conflict. 
And add the other way conversion (int64s to numbers).
Cyberpunk 2077 (C) 2020 by CD Projekt RED 05 05 21 19_36_18
`AssignedIndex` is `Int32`

Sombra0xCC:
@psiberx That would be great
i removed my LuaRED changes

PR #558, opened 2021-05-30
Add ImGui::GetStyle() to sol_imgui.h

Nats-ji:


WSSDude420:
I dont think this would work the way you expect it. + does not look too optimal to be honest if it were to be called a lot.

WSSDude420:
I'm personally against merge till this is rewritten to behave like it should originally. That is, object returned by GetStyle is editable and changes propagate.

You cannot achieve this much more optimally without introducing usertypes.

Nats-ji:
> I'm personally against merge till this is rewritten to behave like it should originally. That is, object returned by GetStyle is editable and changes propagate.
> 
> You cannot achieve this much more optimally without introducing usertypes I believe.

Changed to usertype

WSSDude420 (approved):

WSSDude420 (approved):

PR #561, opened 2021-06-07
Add ResourceAsyncReference support, small fixes and improvements

psiberx:
- `Observe()` and `Override()` now accept both native and scripted type names.
- Improved overloaded functions handling. Prevents constant swapping when two overloads are used at the same rate.
- Improved detection of whether the game is running. Prevents more crashes when exiting the game.
- Added support for old style static calls `obj:Static(obj, args)`. Improves backward compatibility with 1.13 mods.
- Added support for `ResourceAsyncReference`. Allows editing of TweakDB props of that type by the mods,
- Fixed `CName` calculation for "None" (mirrors RED4ext logic).
- Added logging for errors occurred in the module loaded with `require()`. Should make the transition to the new version less painful.
- Added optional params support. For example, `Game.GetTimeSystem():SetTimeDilation("deflect", 0.1, 2.5)`.
- Fixed disappearing footsteps issue. If the mod stores a reference to the player entity, then it must release the reference when a new game session is loaded. Otherwise the issue of disappearing steps will occur.
- Fixed inconsistent `self` and random crashes for overridden functions.

### DIFF
--- a/src/reverse/RTTIHelper.cpp
+++ b/src/reverse/RTTIHelper.cpp
@@ -368,7 +368,7 @@ sol::function RTTIHelper::MakeInvokableFunction(RED4ext::CBaseFunction* apFunc)
     auto& luaState = lockedState.Get();
 
     return MakeSolFunction(luaState, [this, apFunc](sol::variadic_args aArgs, sol::this_environment aEnv) -> sol::variadic_results {
-        uint32_t argOffset = 0;
+        uint64_t argOffset = 0;
         RED4ext::ScriptInstance pHandle = ResolveHandle(apFunc, aArgs, argOffset);
 
         std::string error;
@@ -400,7 +400,7 @@ sol::function RTTIHelper::MakeInvokableOverload(std::map<uint64_t, RED4ext::CBas
         {
             variant->lastError.clear();
 
-            uint32_t argOffset = 0;
+            uint64_t argOffset = 0;
             RED4ext::ScriptInstance pHandle = ResolveHandle(variant->func, aArgs, argOffset);
 
             auto result = ExecuteFunction(variant->func, pHandle, aArgs, argOffset, true, variant->lastError);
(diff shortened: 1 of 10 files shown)


WSSDude420:
If you are still making changes, would be wiser to make it a draft I would say
So it does not get merged by accident x)

psiberx:
No it's complete and can be merged. This last update was unplanned.
And I may not be able to commit footsteps fix anytime soon as the original GC fix didn't work as expected (it fixes the footsteps but can cause other issues for some users).

WSSDude420:
Yeah, read it on Discord, np :)

psiberx:
Sorry, one more unexpected update .) Discovered a real case of in-game var that requires ~5KB, just increased the allocator's limit.

src/reverse/RTTIHelper.cpp
```diff
@@ -520,43 +543,64 @@ sol::variadic_results RTTIHelper::ExecuteFunction(RED4ext::CBaseFunction* apFunc
     ResetAllocator ___allocatorReset;
 
     std::vector<RED4ext::CStackType> callArgs(apFunc->params.size);
```
yamashi, line 545: std::vector should be replaced with TiltedPhoques::Vector, it's an overload that will also use the pushed allocator, so you will get free memory allocations here

yamashi (changes requested):

src/reverse/RTTIHelper.cpp
psiberx: I was using `TiltedPhoques::Vector` when started the RTTI helper and then switched to `std::vector` because in my tests the last one showed surprisingly better performance. Let me do the tests again in the final version.

WSSDude420:
std doesnt suck as much as it used to nowadays x)

Could be tbh 🤷 

Yeah, test should prove it for sure, although hard to test these precisely i guess 😞

yamashi:
> std doesnt suck as much as it used to nowadays x)
> 
> Could be tbh 🤷
> 
> Yeah, test should prove it for sure, although hard to test these precisely i guess 😞

TiltedPhoques::Vector is a std::vector but with a custom allocator to take advantage of the cheap allocations

WSSDude420:
Not saying yours is worse by that x)
I read what you wrote. Was just addition to debate more or less.
I guess yours function more like std::pmr::vector from what you wrote?

src/reverse/RTTIHelper.cpp
psiberx: Fixed.

Issue #562, opened 2021-06-12
Get rid of pre-RED4ext legacy code


WSSDude420:
https://github.com/yamashi/CyberEngineTweaks/blob/dee335c1f1fd612258aaad48131fbdbb09d52430/src/reverse/BasicTypes.h#L67

These should be removed cause maintnance nightmare. Legacy stuff from before RED4ext.

PR #566, opened 2021-06-15
Legacy types begone

WSSDude420:
This PR should fix issue #562

Also fixes;
- ItemID TweakDBID property comparison and assignment


WSSDude420:
Seems to function just as it was till this commit at least... Few peeps have it, I bet they had mods installed with it too (regular users).
No complaints so far, would expect some if something obvious broke.

WSSDude420:
Rebase onto master

WSSDude420:
If anyone wants to pick this up, feel free to do so. Groundwork is here in separate branch...

PR #568, opened 2021-06-24
Xmake install

Nats-ji:
Reopen of https://github.com/yamashi/CyberEngineTweaks/pull/567
Using xmake config instead, as suggested.

Usage
```sh
xmake f --installpath="C:\Program Files (x86)\Steam\steamapps\common\Cyberpunk 2077\bin\x64\plugins"
xmake install
```


yamashi (approved):

yamashi:
Looks good thanks :)

WSSDude420:
This is still not proper install.
Proper install would be to cyberpunk root.

What we use for packaging should be used for this, but in a bit more generic manner (that is not release only for example)
Dunno... But yamashi already made his mind it seems.

WSSDude420:
Mentioning this cause technically, scripts are part of CET (extra folder we package)

If we were to expand those, thoze would too become part of the debugging process and hence shoukd be in install.
And by that point, you could pretty much reuse packaging we have I believe in some way...

Note for an improvement I suppose x)

Nats-ji:
> Mentioning this cause technically, scripts are part of CET (extra folder we package)
> 
> If we were to expand those, thoze would too become part of the debugging process and hence shoukd be in install.
> And by that point, you could pretty much reuse packaging we have I believe in some way...
> 
> Note for an improvement I suppose x)

how about this https://github.com/Nats-ji/CyberEngineTweaks/commit/204ae7bc914da12156a42025c6a7522be9d1064c

waruqi:
why not use `xmake install -o installdir` , set_installdir and get install directory from target:installdir()?

Nats-ji:
> why not use `xmake install -o installdir` or `set_installdir` and get install directory from `target:installdir()` in on_install?

I think you have to type the install directory every time using that? I'm not sure.

PR #628, opened 2021-11-22
Overrides and Variants

psiberx:
1. Improved function overrides.
- When using `Override()` the handler receives the original function (or next in the override chain) as the last parameter
```lua
Override('PlayerPuppet', 'SetWarningMessage', function(_, message, wrappedMethod)
    wrappedMethod('[Overridden] ' .. message)
end)
```
- The wrapped function can be called multiple times
```lua
Override('PlayerPuppet', 'GetGunshotRange', function(_, originalRangeFunc)
    return originalRangeFunc() + originalRangeFunc()
end)
```
- If any of the overrides fail then the entire override chain also fails
- Observers can be called before and after the target function using `ObserveBefore()` and `ObserveAfter()`
```lua
ObserveBefore('PlayerPuppet', 'GetGunshotRange', function()
    print('PlayerPuppet::GetGunshotRange::Before')
end)
ObserveAfter('PlayerPuppet', 'GetGunshotRange', function()
    print('PlayerPuppet::GetGunshotRange::After')
end)
```
- `Observe()` is an alias for `ObserveBefore()`
- All observers are guaranteed to be called even if overrides and other observers fail
- It's possible to observe and override static class methods
```lua
Override('RPGManager', 'GetFloatItemQuality', function(value)
    print('RPGManager.GetFloatItemQuality(' .. value .. ')')
    return gamedataQuality.Legendary
end)
```
- Full name should be used when overriding a scripted static method
```lua
Observe('PlayerPuppet', 'IsSwimming;PlayerPuppet', function(player)
    print('PlayerPuppet.IsSwimming(player)')
end)
```
2. Improved `Variant` type support.
- Use `FromVariant(variant)` to extract the value 
- Use `ToVariant(value)` to pack the value when the type is unambiguous
```lua
ToVariant(Vector4.new(1, 2, 3))
ToVariant(gamedataQuality.Epic)
ToVariant("Nam libero tempore")
ToVariant(16570246047455160070ULL)
ToVariant(true)
ToVariant(TweakDBID("Items.money"))
ToVariant(CName("deflect"))
ToVariant(LocKey(12345))
ToVariant(CRUID(1337))
```
- Use `ToVariant(value, type)` when the type is ambiguous or you wanna force the type
```lua
ToVariant(1, "Int32") -- Same numeric value can be a signed integer
ToVariant(1, "Uint32") -- or unsigned integer
ToVariant(1, "Float") -- or float
ToVariant(nil, "handle:PlayerPuppet") -- Null requires a concrete type
```
- The type of the array is resolved from the first element when possible
```lua
ToVariant({ "a", "b", "c" }) -- Auto resolved as "array:String"
ToVariant({ 1, 2, 3 }, "array:Uint32") -- Ambiguous type
```
3. Improved `ISerializable` types support. They are now correctly wrapped in `ref<>` when created.

### DIFF
--- a/src/reverse/BasicTypes.cpp
+++ b/src/reverse/BasicTypes.cpp
@@ -135,7 +135,7 @@ Variant::~Variant()
 
 bool Variant::IsEmpty() const noexcept
 {
-    return type == nullptr;
+    return !type;
 }
 
 bool Variant::IsInlined() const noexcept
@@ -150,12 +150,12 @@ RED4ext::CBaseRTTIType* Variant::GetType() const noexcept
 
 RED4ext::ScriptInstance Variant::GetDataPtr() const noexcept
 {
-    return IsInlined() ? (RED4ext::ScriptInstance)inlined : instance;
+    return IsInlined() ? const_cast<uint8_t*>(inlined) : instance;
 }
 
 bool Variant::Init(const RED4ext::CBaseRTTIType* aType)
 {
-    if (aType == nullptr)
+    if (!aType)
     {
         Free();
         return false;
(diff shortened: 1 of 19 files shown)


yamashi:
Thanks, this is awesome!

PR #629, opened 2021-11-25
Overrides fixes, TweakDB hook

psiberx:
1. Use weak handle for `self` in overrides. This should decrease the number of unreleased references and make CET less dependent on garbage collection.

2. Reuse overridden functions on mods reload. This fixes a potential issue with the native function registry, which has a hard limit on the number of registered functions.

3. Support out params in overrides.  
To properly handle out params in overrides:  
- Accept all params in the handler
- Don't pass out params to the wrapped function
- Return all results from the handler
```lua
registerForEvent('onInit', function()
    Override('SpatialQueriesSystem', 'SyncRaycastByCollisionGroup', function(_, a1, a2, a3, a4, a5, a6, wrapped)
        -- Skip a4 here because it's an out param
        local success, result = wrapped(a1, a2, a3, a5, a6)
        print(success, GameDump(result))

        -- Return all results
        return success, result
    end)
end)

registerHotkey('SyncRaycastByCollisionGroup', 'SyncRaycastByCollisionGroup', function()
    local from, forward = Game.GetTargetingSystem():GetCrosshairData(GetPlayer())
    local to = Vector4.new(from.x + forward.x * 100, from.y + forward.y * 100, from.z + forward.z * 100, from.w)

    local success, result = Game.GetSpatialQueriesSystem():SyncRaycastByCollisionGroup(from, to, 'Static', false, false)
    print(success, GameDump(result))
end)
```

4. Added event for early TweakDB changes:
```lua
registerForEvent('onTweak', function()
	TweakDB:SetFlat('PreventionSystem.setup.totalEntitiesLimit', 20)
end)
```

### DIFF
--- a/src/scripting/Scripting.cpp
+++ b/src/scripting/Scripting.cpp
@@ -126,7 +126,7 @@ void Scripting::Initialize()
     m_store.LoadAll();
 }
 
-void Scripting::PostInitialize()
+void Scripting::PostInitializeStage1()
 {
     auto lua = m_lua.Lock();
     auto& luaVm = lua.Get();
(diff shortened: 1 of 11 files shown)


src/scripting/Scripting.cpp
```diff
@@ -135,8 +135,7 @@ void Scripting::PostInitializeStage1()
 
     luaVm.new_usertype<Scripting>("__Game",
         sol::meta_function::construct, sol::no_constructor,
-        sol::meta_function::index, &Scripting::Index,
-        sol::meta_function::new_index, &Scripting::NewIndex);
```
yamashi, line 139: If new_index is removed, you can just remove InternalNewIndex?

psiberx: Yeah, we can remove it.

psiberx: And InternalIndex too actually, it's used in Index only.

yamashi (approved):

PR #697, opened 2022-09-11
TweakDBEditor: Add support for REDMod modified TweakDB

dragonzkiller:
Adds support for reading and importing entries from the TweakDB created by REDMod.

The `HasREDModTweakDB` function isn't really needed since you can just check the file directly and doesn't have to be public if it's kept, but maybe it's good to have it in there if there's a need to check later. Not sure where though. Maybe it could be used to update the TweakDB Editor window to reflect that CET found a modified TweakDB. But if that's the case it might be better to move it to a function that checks that it was properly loaded like how `IsInitialized` is used.

Let me know what you think.

### DIFF
--- a/src/Paths.cpp
+++ b/src/Paths.cpp
@@ -35,6 +35,11 @@ const std::filesystem::path& Paths::ArchiveModsRoot() const
     return m_archiveModsRoot; 
 }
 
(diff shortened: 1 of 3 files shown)


yamashi (approved):
Looks good to me!

PR #704, opened 2022-09-25
Scripting

psiberx:
- Fixed deadlocks for some overrides combinations
- Added support for scripts hot reload

### DIFF
--- a/src/CET.cpp
+++ b/src/CET.cpp
@@ -61,7 +61,7 @@ bool CET::IsRunning() noexcept
 CET::CET()
     : m_options(m_paths)
     , m_bindings(m_paths, m_options)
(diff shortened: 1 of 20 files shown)


src/reverse/RTTIMapper.cpp
```diff
@@ -130,55 +141,60 @@ void RTTIMapper::RegisterSimpleTypes(sol::state& aLuaState, sol::table& aLuaGlob
         data.value = aVariant.GetDataPtr();
 
         auto luaLock = m_lua.Lock();
-        
+
```
tabaraemanuel, line 144: Remove this ?

WSSDude420: i dont see an issue with removing trailing whitespace honestly, unless it makes changes unreadable

PR #707, opened 2022-09-28
Overlay: (Windows) Fixed cyrillic fonts incorrect drawing

ForserX:


WSSDude420:
You can set font and glyphs used from options json already. If you want it to be automatic to some degree, do it properly and not like this, hardcoding it for specific language. This is not a good solution.

See https://github.com/yamashi/CyberEngineTweaks/blob/0e10d09365dfd6b9a6d24d46f59cbab295b6a98f/src/d3d12/D3D12_Functions.cpp#L308

Options are font_glyph_ranges, font_path and font_size (latter option unrelated to this in particular)

ForserX:
Config? Why not just optimize for the language of the system?

WSSDude420:
I just mentioned you can currently configure it at least.

I have nothing against having it automatically sleect glyph range or something, I have issue with hardcoded bulgarian constants not improving anything. Your solution is good only for single language on top of that.

If you rework this to be more general, without use of hardcoded values, I think nobody will bet an eye :)

ForserX:
I'm almost done. The only thing I do not know what to do about the Ukrainian encoding. There's a kind of cross between Cyrillic and Latin.

PR #708, opened 2022-09-28
Auto detect system glyph ranges by system lang

ForserX:


WSSDude420:
This looks much better in my opinion 👍 

Just add that fallthrough and if pipeline runs, I think this could be merged.

ForserX:
> Just add that fallthrough and if pipeline runs
Can I go into more detail?

WSSDude420 (approved):

PR #714, opened 2022-10-02
TweakDB, Overrides, RedHotScripts

psiberx:
- Updated TweakDB implementation (faster initialization, flat reading, record creations, better compatibility with other TweakDB tools)
- Fixed deadlocks for some overrides combinations
- Added compatibility with scripts hot reload


WSSDude420:
Merged

Issue #700, opened 2022-09-17
Feature request


Neurolinked:
It will be possible to add or adapt the function **ModArchiveExists** to verify for _file_.archive in the new **redmod** folder

WSSDude420:
Closed by merge of #721

PR #721, opened 2022-10-07
ModArchiveExists: Enable REDmod archive support

dragonzkiller:
Implements #700 

Enables support for searching for archives with `ModArchiveExists` in the REDmod mods directory.

The function first checks the legacy folder, then recursively searches the REDmod directory. It then verifies that the base folder is named `archives` and is two folders deep (since we won't know the mod's name).

This is missing the relative path check like with `cRelPathStr`, but it should be easy enough to add if it's required.


WSSDude420:
Please, update it to use GetLuaPath (see latest commit)

For some examples of how it could be setup with it, look into for example `src\scripting\LuaSandbox.cpp` into `os` function implementations.

src/scripting/Scripting.cpp
```diff
@@ -97,13 +97,44 @@ void Scripting::Initialize()
 
     luaGlobal["ModArchiveExists"] = [this](const std::string& acArchiveName) -> bool
     {
+        bool archiveExists = false;
```
WSSDude420, line 100: Update PR to latest master and use GetLuaPath instead of manually managing paths like this ideally.

src/scripting/Scripting.cpp
```diff
@@ -97,13 +97,44 @@ void Scripting::Initialize()
...
+            archiveExists = exists(cAbsPath);
+
+        // only need to check if it wasn't already found
+        if (!archiveExists)
+        {
+            // check to see if a REDmod archive exists
+            const auto cREDModAbsPathRoot = absolute(m_paths.REDmodsRoot());
+
+            // parse recursively to find the archive
+            for (auto dir_iter = std::filesystem::recursive_directory_iterator{cREDModAbsPathRoot};
```
WSSDude420, line 117: I would not use recursive_directory_iterator when you care about specific level only. 
Also not a fan of this type of iteration this looks better IMO.
Lastly, as mentioned before, use GetLuaPath function for validating things please.
```
for (const auto& redModIter : std::filesystem::directory_iterator(cREDModsAbsPathRoot))
{
    if (!redModIter.is_directory())
        continue;
        
    for (const auto& entry : std::filesystem::directory_iterator(redModIter.path()))
    {
        if (!entry.is_directory())
            continue;
            
        // check path, again using GetLuaPath
    }
}
```

WSSDude420 (changes requested):

src/scripting/Scripting.cpp
dragonzkiller: no longer valid due to RED4ext::ResourceDepot implementation

src/scripting/Scripting.cpp
dragonzkiller: no longer valid due to RED4ext::ResourceDepot implementation

dragonzkiller:
Converted the entire function to use `RED4ext::ResourceDepot` from @psiberx 

Confirmed to work in all three cases:
1.) legacy `archive\pc\mod`: found
2.) enabled & deployed REDmod mod: found
3.) disabled REDmod mod: not found

src/reverse/RED4ext/ResourceDepot.hpp
```diff
@@ -0,0 +1,65 @@
+#pragma once
```
WSSDude420, line 1: Put this header in src/reverse, no need for new subfolder

src/scripting/Scripting.cpp
```diff
@@ -24,6 +24,8 @@
 #include <reverse/RTTIExtender.h>
 #include <reverse/Converter.h>
```
WSSDude420, line 26: Get rid of new line

src/reverse/RED4ext/ResourceDepot.hpp
WSSDude420: Also, we use .h for headers, would not mix header extensions rather

xmake.lua
```diff
@@ -42,7 +42,7 @@ target("cyber_engine_tweaks")
     set_kind("shared")
     set_filename("cyber_engine_tweaks.asi")
     add_files("src/**.cpp")
-    add_headerfiles("src/**.h", "build/CETVersion.h")
+    add_headerfiles("src/**.h", "src/**.hpp", "build/CETVersion.h")
```
WSSDude420, line 45: When you change extension, you can remove this change

WSSDude420 (changes requested):

src/reverse/RED4ext/ResourceDepot.hpp
dragonzkiller: Renamed file

xmake.lua
dragonzkiller: Removed ".hpp" references

dragonzkiller:
Sorry for the number of commits as I'm away from my computer right now.  But that should resolve the requested changes. Let me know if you find anything else.

PR #751, opened 2022-11-14
Fix possible deadlock on startup

WSSDude420:
If user toggles overlay during early startup, focuses on console input and presses overlay toggle hotkey, deadlock happens.

ImGui got locked in main thread and render thread due to some ImGui things...

We do not access platform things from present now, modified a bit log there.
Removed lock of ImGui in OnWndProc.

Reworked GameMainThread class as part of this, used it for managing tasks for each state type tick function.
Currently used in LuaVM and D3D12, along with dump job.

Dependency for https://github.com/yamashi/CyberEngineTweaks/pull/754

### DIFF
--- a/src/scripting/GameDump.cpp
+++ b/src/scripting/GameDump.cpp
@@ -4,7 +4,7 @@
 
 namespace GameDump
 {
-void DumpVTablesTask::Run()
+bool DumpVTablesTask::Run()
 {
     TiltedPhoques::Map<uintptr_t, std::string> vtableMap;
 
@@ -82,5 +82,7 @@ void DumpVTablesTask::Run()
     {
         Log::Info("{:016X},{}", key, value);
     }
+
+    return true;
 }
 } // namespace GameDump
--- a/src/scripting/GameHooks.cpp
+++ b/src/scripting/GameHooks.cpp
@@ -6,13 +6,13 @@
 
 static std::unique_ptr<GameMainThread> s_pGameMainThread;
 
-void GameMainThread::TaskQueue::AddTask(const std::function<bool()>& aFunction)
+void GameMainThread::RepeatedTaskQueue::AddTask(const std::function<bool()>& aFunction)
 {
     std::lock_guard lock(m_mutex);
     m_tasks.emplace_back(aFunction);
 }
 
-void GameMainThread::TaskQueue::Drain()
+void GameMainThread::RepeatedTaskQueue::Drain()
 {
     std::lock_guard lock(m_mutex);
     for (auto taskIt = m_tasks.begin(); taskIt != m_tasks.end();)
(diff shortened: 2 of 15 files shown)


WSSDude420:
Realized later that LuaVM hook is called after scripting system is enabled...

Moved it on GameMainThread task.

WSSDude420:
Was not, should be now... 

Reworked GameMainThread a bit, allows to schedule tasks for all state tick functions independently.
Moved things which were previously hooking these as jobs there (LuaVM, D3D12, RTTI dump)

Interface should be somewhat usable for future hopefully.

src/scripting/GameHooks.cpp
```diff
@@ -37,34 +60,123 @@ GameMainThread::~GameMainThread()
     Unhook();
 }
 
-bool GameMainThread::HookMainThread(void* a1, void* a2)
+bool GameMainThread::HookMainThreadStateTick(RED4ext::IGameState* apThisState, RED4ext::CGameApplication* apGameApplication)
 {
     auto& gmt = Get();
 
-    gmt.m_taskQueue.Drain();
+    for (auto taskIt = gmt.m_genericTasks.begin(); taskIt != gmt.m_genericTasks.end(); ++taskIt)
```
yamashi, line 67: Mistake here ++taskIt should be in an else condition

src/scripting/GameHooks.h
```diff
@@ -7,22 +7,29 @@ struct GameMainThread
...
 
 private:
     GameMainThread();
-    
+
     void Hook();
     void Unhook() const;
 
-    using TMainThreadLoop = bool(void*, void*);
+    using TMainThreadStateTick = bool(RED4ext::IGameState*, RED4ext::CGameApplication*);
```
yamashi, line 24: A struct should be more appropriate here

src/scripting/GameHooks.cpp
```diff
@@ -19,12 +21,33 @@ void GameMainThread::Shutdown()
...
 GameMainThread& GameMainThread::Get()
 {
+    Initialize();
     return *s_pGameMainThread;
 }
 
-void GameMainThread::AddTask(std::function<void()> aFunction)
+void GameMainThread::AddBaseInitializationTask(const std::function<bool()>& aFunction)
+{
+    m_baseInitializationTasks.emplace_back(aFunction);
```
yamashi, line 30: All of these aren't thread safe

yamashi (changes requested):

src/scripting/GameDump.cpp
```diff
@@ -82,5 +82,7 @@ void DumpVTablesTask::Run()
     {
         Log::Info("{:016X},{}", key, value);
     }
+
+    return true;
```
yamashi, line 86: What's the point of this return ?

WSSDude420: this was mentioned in commit message
> - can autodestruct task when task says it is finished (returns true)

WSSDude420: it is so it does not repeat it

src/scripting/GameHooks.h
WSSDude420: dont know what you mean by this exactly, changed the rest 🤔

yamashi (approved):

src/scripting/GameHooks.h
yamashi: don't use std::pair, first and second are not expressive, it's hard to understand what they are

WSSDude420: oh, this one
kk, i got confused cause it shows this comment on another line 😄

WSSDude420: cleaned it up, think this looks rather nice now 👍

WSSDude420: actually, wait a bit - want to do one more thing

WSSDude420: k think this is really done now 😁

yamashi (approved):

PR #752, opened 2022-11-14
Persistent state for Overlay widget toggles

WSSDude420:
Keeps previously opened widgets opened after game shuts down when it next launches.

Also organize options a bit better and fix mod being able to draw during first time setup.

Dependency for https://github.com/yamashi/CyberEngineTweaks/pull/754
Continued in #754 (gave it separate file there and some other tweaks)

### DIFF
--- a/src/Options.cpp
+++ b/src/Options.cpp
@@ -3,106 +3,161 @@
 #include "Paths.h"
 #include "Utils.h"
 
+
+void PatchesSettings::Load(const nlohmann::json& aConfig)
+{
+    RemovePedestrians = aConfig.value("remove_pedestrians", RemovePedestrians);
+    SkipStartMenu = aConfig.value("skip_start_menu", SkipStartMenu);
+    AmdSmt = aConfig.value("amd_smt", AmdSmt);
+    AsyncCompute = aConfig.value("disable_async_compute", AsyncCompute);
+    Antialiasing = aConfig.value("disable_antialiasing", Antialiasing);
+    DisableIntroMovies = aConfig.value("disable_intro_movies", DisableIntroMovies);
+    DisableVignette = aConfig.value("disable_vignette", DisableVignette);
+    DisableBoundaryTeleport = aConfig.value("disable_boundary_teleport", DisableBoundaryTeleport);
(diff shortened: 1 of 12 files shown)


src/Options.cpp
```diff
@@ -3,106 +3,206 @@
...
+      {"disable_intro_movies", DisableIntroMovies},
+      {"disable_vignette", DisableVignette},
+      {"disable_boundary_teleport", DisableBoundaryTeleport},
+      {"disable_win7_vsync", DisableWin7Vsync},
+      {"minimap_flicker", MinimapFlicker},
+    };
 }
 
-void Options::Save()
+void PatchesSettings::ResetToDefaults()
```
yamashi, line 37: Could just write *this = {};

src/Options.cpp
```diff
@@ -3,106 +3,206 @@
...
+      {"oversample_vertical", OversampleVertical}
+    };
+}
 
-    config["font_path"] = FontPath;
-    config["font_glyph_ranges"] = FontGlyphRanges;
-    config["font_size_base"] = FontSizeBase;
-    config["font_oversample_horizontal"] = FontOversampleHorizontal;
-    config["font_oversample_vertical"] = FontOversampleVertical;
+void FontSettings::ResetToDefaults()
```
yamashi, line 71: *this = {};

src/Options.cpp
```diff
@@ -3,106 +3,206 @@
...
+    return {
+      {"remove_dead_bindings", RemoveDeadBindings},
+      {"enable_imgui_assertions", EnableImGuiAssertions},
+      {"enable_debug", EnableDebug},
+      {"dump_game_options", DumpGameOptions},
+    };
+}
 
+void DeveloperSettings::ResetToDefaults()
+{
```
yamashi, line 105: *this = {}

yamashi (changes requested):

src/Options.cpp
WSSDude420: did not thought of that 💡

WSSDude420: changed all of these

yamashi (approved):

PR #753, opened 2022-11-14
Better logging

WSSDude:
Idea is to give mods better logging facilities to both, console and their own logs (which they now only have in form of a file).

This should enable colored console output and allow mods to enable their own logging window at will, which would contain only their log. Logging into these sinks is provided via new consoleLog and modLog functions.

consoleLog works also in console, is meant as a replacement for print (print is still supported and behaves as consoleLog.info)
modLog is mod-specific, points to mod-specific logger and log window (aliases spdlog for backwards compatibility)

Both of these new interfaces support printing of multiple arguments (converts them to strings and concats them like print did previously).
This is upgrade over previous spdlog implementation, which it inherits thanks to shared interface.

Conflicts with PR #757 (Console.cpp)

consoleLog
modLog

### DIFF
--- a/src/CET.cpp
+++ b/src/CET.cpp
@@ -71,7 +71,7 @@ CET::CET()
     , m_bindings(m_paths, m_options)
     , m_window(&m_bindings, &m_d3d12)
     , m_d3d12(m_window, m_paths, m_options)
-    , m_vm(m_paths, m_bindings, m_d3d12)
+    , m_vm(m_paths, m_options, m_bindings, m_d3d12)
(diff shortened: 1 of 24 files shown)


WSSDude420:
Added images from Discord I made earlier.

WSSDude420:
Merged log sizes into this and fixed compilation.

WSSDude420:
From test merge of all of these PRs, this one is colliding with #758.

Whichever gets merged second needs to be rebased.

PR #754, opened 2022-11-14
WIP: Better persistency and limit log and history sizes

WSSDude420:
- Persistent state in its own class (and file)
- Persistent console history
- Configurable maximum number of cached lines in memory for each log and history
- If someone doesnt want to save console history persistently, there is option for this too

Requires ~~#751~~, ~~#752~~ and #753
Needs to be rebased after they are merged


WSSDude420:
Redundant, created #758 and merged the rest into #753

PR #759, opened 2022-11-18
Update ImGui backend impls to v1.89

WSSDude420:
- update ImGui to v1.89 docking (needs to be merged into xmake-repo https://github.com/xmake-io/xmake-repo/pull/1639 )
- update ImGui backends to corresponding ImGui version
- update enables docking with minimal changes
- enables proper DPI scaling

Conflicts with PR #757, there will need to be rebase.


WSSDude420:
Ported changes we did to backends.

Backends here were directly copied from 1.89 docking release. That extra commit introduces our fixes back again with minimal changes.

I would rather keep this file as close to reference as possible, we should have not previously refactored it...

PR #762, opened 2022-11-28
Improve RTTI calls

psiberx:
Building `CStackFrame` and bytecode for arguments on our side now, which allows to:
- Pass optional params as intended
- Call a couple more functions

### DIFF
--- a/src/reverse/RTTIHelper.cpp
+++ b/src/reverse/RTTIHelper.cpp
@@ -10,11 +10,19 @@
 #include <scripting/Scripting.h>
 #include <Utils.h>
 
-static constexpr bool s_cEnableOverloads = true;
-static constexpr bool s_cLogAllOverloadVariants = true;
-static constexpr bool s_cThrowLuaErrors = true;
+namespace
+{
+constexpr bool s_cEnableOverloads = true;
+constexpr bool s_cLogAllOverloadVariants = true;
+constexpr bool s_cThrowLuaErrors = true;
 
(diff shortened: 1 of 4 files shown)


WSSDude420 (approved):

src/reverse/RTTIHelper.cpp
```diff
@@ -770,6 +771,53 @@ sol::variadic_results RTTIHelper::ExecuteFunction(RED4ext::CBaseFunction* apFunc
...
+    constexpr auto PointerSize = sizeof(void*);
+
+    char code[MaxCodeSize];
+    RED4ext::CStackFrame frame(nullptr, code);
+
+    for (uint32_t i = 0; i < apFunc->params.size; ++i)
+    {
+        const auto& param = apFunc->params[i];
+        const auto& arg = aArgs[i];
+
```
WopsS, line 790: I would add a check for current `frame.code` pointer, if someone supplies too many args it will overflow the buffer.

Also, do you think it is a good idea to port this to the SDK?

psiberx: Yeah, I think it makes sense to add to SDK. And if we want to make the function calling safer, then we have to add checks to all methods, since the engine doesn't check anything. This is already safer, because it relies on the actual params from RTTI, not on passed args. Whereas the original `ExecuteNative` just writes all the args you pass with `CStack` without even checking what the function accepts and without checking the buffer length.

PR #765, opened 2022-12-01
Add icon font support

Nats-ji:
https://github.com/yamashi/CyberEngineTweaks/issues/755

```lua
ImGui.Text("icons: "..IconGlyphs.Music..IconGlyphs.GlassCocktail..IconGlyphs.SearchWeb..IconGlyphs.Cog..IconGlyphs.Github..IconGlyphs.Facebook)
```
image

Update:
switched to [material design icons](https://github.com/Templarian/MaterialDesign) which has over 7k icons: https://materialdesignicons.com/


src/scripting/Scripting.cpp
```diff
@@ -59,6 +59,7 @@ void Scripting::Initialize()
     const auto previousCurrentPath = std::filesystem::current_path();
     current_path(m_paths.CETRoot() / "scripts");
     luaVm.script("json = require 'json/json'", sol:: detail::default_chunk_name(), sol::load_mode::text);
+    luaVm.script("FA = require 'fontawesome/icons'", sol:: detail::default_chunk_name(), sol::load_mode::text);
```
WSSDude420, line 62: I would probably name it IconGlyphs or something, is more expressive than FA

WSSDude420 (changes requested):
Please, use latest version 6 if possible, and desktop version ideally instead of web one.

scripts/fontawesome/icons.lua
```diff
@@ -0,0 +1,679 @@
+local icons = {
+    ICON_GLASS = '\u{f000}',
```
WSSDude420, line 2: I dont know if you generated this somehow or where did you get this, but if you rename the enum like I wrote, remove the ICON_ prefix here please (so it doesnt look silly... IconGlyphs.ICON_GLASS is kinda weird)

WSSDude420: Bonus points if you could make it like 

ICON_GLASS -> Glass
ICON_TH_LARGE -> ThLarge

so it fits more with a style of our other enums

WSSDude420:
After looking around git a bit, couldn't we utilize this in some way?

https://github.com/juliettef/IconFontCppHeaders

This way, we could use it from C++ also (would like to make use of the icons in CET itself probably at some point)

Nats-ji:
> After looking around git a bit, couldn't we utilize this in some way?
> 
> https://github.com/juliettef/IconFontCppHeaders
> 
> This way, we could use it from C++ also (would like to make use of the icons in CET itself probably at some point)

not sure where to put it, so I just added it to the vendor folder.

> Please, use latest version 6 if possible

Let's just use version 4 since version 5/6 glyph range conflicts between light and solid. 😉

WSSDude420:
Create new xmake package
https://github.com/xmake-io/xmake-repo

Vendor is very specific folder, would not add more stuff to it...

If you dont know how to do xmake package, can have a look later... Cant find time for CET in past few fays :(

WSSDude420:
And in regards to FA6, would load the fonts from it in following order:
- brands
- regular
- solid

WSSDude420:
In regards to FA6, would load in following order
- regular
- solid
- brands

maybe even skipped regular icons, looks like solid has them all (and they should look the best hopefully in current UI)

Nats-ji:
switched to font to https://materialdesignicons.com/

WSSDude420 (approved):

PR #779, opened 2022-12-31
💅🏻 Add Simple Theme Load/Dump from JSON

effs:
- [x] `cet/theme.json` loaded if it's present and valid
- [x] Any (0..N) known ImGuiStyle attributes and colors are used from the theme
- [x] Effective theme dumped to log
- [x] `GetScaleFactor()` exposes current UI scaling factor through Lua
  - [x] ~Maybe better name?~
- [x] Specify all supported attributes/colors (whether all or a restricted set) as a map for some degree of control 
- [x] Default CET theme installed as `<cet>/theme.json` 

Maybe?
- [ ] Allow dumping ImGuiDebug-defined theme?
- [ ] Allow writing theme to file?
- [ ] Allow supplying a global scaling override for UI/fonts?

➕ Adds [fmt](https://github.com/fmtlib/fmt) and [scnlib](https://github.com/eliaskosunen/scnlib) as dependencies (`fmt` is the impl of `std::format`, `scnlib` the impl for parsing equivalent proposal)

⚡ Doesn't introduce any new mechanisms or tools within CET or for CET moddevs, so compatibility issues with any future theming plans should be absolutely minimal.

Example theme.json:
```json
{
    "ImGuiThemeVersion": "1.0.0",
    "ThemeName": "Straight outta JSON",
    "style": {
        "WindowRounding": 7.0
    },
    "colors": {
        "Text": "#ff111199",
        "WindowBg": "#11ff11"
    }
}
```

Or the current CET defaults would look like:
```json
{
    "ImGuiThemeVersion": "1.0.0",
    "ThemeName": "CET Default",
    "style": {
        "WindowRounding": 7.0,
        "WindowTitleAlign": [0.5, 0.5],
        "ChildRounding": 6.0,
        "PopupRounding": 6.0,
        "FrameRounding": 6.0,
        "ScrollbarRounding": 12.0,
        "GrabRounding": 12.0,
        "TabRounding": 6.0
    }
}
```


effs:
The codec creation can be dried up a bit, and I think there's a way to at least validate we've got the right kind of codec for the field, and maybe some absurd compile-time way to validate that we really have all fields covered... but it works, I think.

yamashi (changes requested):
Run clang-format and give more explicit names to the various functions, for example "XY" is not explicit.

Other than that, looks good to me :)

vendor/xmake-local/packages/s/scnlib/xmake.lua
```diff
@@ -0,0 +1,40 @@
+package("scnlib")
```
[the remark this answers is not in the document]
effs: [Did and it's been merged](https://github.com/xmake-io/xmake-repo/blob/master/packages/s/scnlib/xmake.lua), although not available to pull yet.. I'll follow up on that.

src/d3d12/D3D12.h
```diff
@@ -15,6 +15,8 @@ struct D3D12
 {
     inline static const uint32_t g_numDownlevelBackbuffersRequired = 3; // Windows 7 only: number of buffers needed before we start rendering
 
+    inline static const SIZE g_designResolution = { 1920, 1080 };
```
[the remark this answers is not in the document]
effs: Hm, how would you structure it that way? I used `SIZE` because it's what the actual resolution value uses - so this is literally `1920x1080` (and I guess you could provide multiple to select from or allow the user to supply their own reference resolution.)

In an enum the x and y would be divorced from one another? So my typing sense would make it something like

```c++
const enum Resolutions {
  DefaultDesignResX = 1920,
  DefaultDesignResY = 1080
}
```
Am I missing a different way to use enums here?

vendor/xmake-local/packages/s/scnlib/xmake.lua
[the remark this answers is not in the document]
effs: Pulling from the main repo now.

effs:
- [x] The clang is formatted
- [x] Switched from `XY` to `XyDimensions` (better ideas welcome)
- [x] `scnlib` pulled from main repo, nixed the local

- [ ] Couldn't figure out a good way to use an enum for the design resolution

- ➕ Added default CET theme as the default `theme.json` installed in the CET dir

src/Theme.cpp
```diff
@@ -0,0 +1,413 @@
...
+    return true;
+}
+
+//
+// API
+//
+
+bool LoadStyleFromThemeJson(const std::filesystem::path& aThemePath, ImGuiStyle& aOutStyle)
+{
+    bool loaded = false;
```
yamashi, line 381: this variable is not required

src/Theme.cpp
```diff
@@ -0,0 +1,413 @@
...
+            Log::Info("No theme file found!");
+        }
+        else
+        {
+            loaded = StyleFromThemeJson(aOutStyle, themeJson);
+            if (!loaded)
+                Log::Warn("Loading theme failed!");
+        }
+
+        themeJson.close();
```
yamashi, line 401: this is called by ifstream's destructor

yamashi (changes requested):

src/Theme.cpp
effs: There's 3 failure returns and 1 success return, this seemed like a convenient way to just have a single default-fail return statement. Would there be a better way to do the same thing?

xmake.lua
```diff
@@ -67,7 +69,7 @@ target("cyber_engine_tweaks")
     add_headerfiles("src/**.h", "build/CETVersion.h")
     add_includedirs("src/", "build/")
     add_syslinks("User32", "Version", "d3d11")
-    add_packages("spdlog", "nlohmann_json", "minhook", "hopscotch-map", "imgui", "mem", "sol2", "tiltedcore", "sqlite3", "openrestry-luajit", "xbyak", "stb")
+    add_packages("spdlog", "nlohmann_json", "minhook", "hopscotch-map", "imgui", "mem", "sol2", "tiltedcore", "sqlite3", "openrestry-luajit", "xbyak", "stb", "fmt", "scnlib")
```
WSSDude, line 72: We use C++20, fmt is just std::format, nothing more. + it should already be included (in case someone doesnt use C++20) by spdlog. I have a feel it is redundant to include separately.

src/reverse/WeakReference.cpp
```diff
@@ -1,4 +1,4 @@
-#include <stdafx.h>
+#include "stdafx.h"
```
WSSDude, line 1: why did you change all of these? please, revert these. it may make some sense if it is in src folder, but definitely not in reverse sub folder. relative path is ../stdafx.h then.
i used <> everywhere for consistency sake tbh, doesnt have to be like that in every file... but this is misleading someone reading the code.

WSSDude: and it just makes this a bit hard to review tbh, just revert these please.

WSSDude (changes requested):

src/Theme.cpp
WSSDude: you may explicitly return false. also, json file should get closed automatically when variable leaves the scope. it should simplify the logic a bit if you incorporate these. i believe that is what yamashi meant.

src/Theme.cpp
```diff
@@ -0,0 +1,411 @@
...
+        return ImGuiDir_Up;
+    if (aDirectionLabel == "Down")
+        return ImGuiDir_Down;
+
+    return aDefaultDirection;
+}
+
+static std::string ColorToHex(const ImVec4& aColor)
+{
+    return fmt::format(
```
WSSDude, line 55: as mentioned before, prefer std::format instead of fmt::format everywhere

src/Theme.cpp
```diff
@@ -0,0 +1,411 @@
...
+static ImVec4 ColorFromHex(const std::string& aHex, const ImVec4& aDefaultColor)
+{
+    if (aHex.empty())
+    {
+        return aDefaultColor;
+    }
+
+    int r, g, b = 0;
+
+    auto foundRGB = scn::scan(aHex, "#{:2x}{:2x}{:2x}", r, g, b);
```
WSSDude, line 69: const auto cFoundRGB

src/Theme.cpp
```diff
@@ -0,0 +1,411 @@
...
+}
+
+static ImVec4 ColorFromHex(const std::string& aHex, const ImVec4& aDefaultColor)
+{
+    if (aHex.empty())
+    {
+        return aDefaultColor;
+    }
+
+    int r, g, b = 0;
```
WSSDude, line 67: prefer cstdint types, try to avoid int short long etc.
i see bunch of these, please change these to int32_t int16_t etc.

src/Theme.cpp
```diff
@@ -0,0 +1,411 @@
...
+    std::string valueName;
+    EncodeValueFunc encodeFrom;
+    DecodeValueFunc decodeInto;
+};
+
+//
+// Simple scalar types
+//
+
+template <typename TAccessor> static EncodeValueFunc EncodeScalar(std::string aValueName, TAccessor aValueAccessor)
```
WSSDude, line 106: const std::string& aValueName
dont unnecessarily copy
= in lambda is probably fine as you want to make a copy there, but you do 2 instead of 1 this way

src/Theme.cpp
```diff
@@ -0,0 +1,411 @@
...
+{
+    return [=](const ImGuiStyle& aStyle, json& aOutJson) -> void
+    {
+        aOutJson[aValueName] = aStyle.*aValueAccessor;
+    };
+}
+
+template <typename TAccessor> static DecodeValueFunc DecodeScalar(std::string aValueName, TAccessor aValueAccessor)
```
WSSDude, line 114: https://github.com/yamashi/CyberEngineTweaks/pull/779/files#r1118094707

src/Theme.cpp
```diff
@@ -0,0 +1,411 @@
...
+{
+    return [=](ImGuiStyle& aOutStyle, const json& aJson) -> void
+    {
+        aOutStyle.*aValueAccessor = aJson.value(aValueName, aOutStyle.*aValueAccessor);
+    };
+}
+
+template <typename TAccessor> static Codec Scalar(std::string aValueName, TAccessor aValueAccessor)
```
WSSDude, line 122: ok, i see this bunch of times... probably better if you went over it yourself and change all of these instances whenever you can so whole review does not revolve around this...

src/Theme.cpp
```diff
@@ -0,0 +1,411 @@
...
+    Color("DragDropTarget", ImGuiCol_DragDropTarget),
+    Color("NavHighlight", ImGuiCol_NavHighlight),
+    Color("NavWindowingHighlight", ImGuiCol_NavWindowingHighlight),
+    Color("NavWindowingDimBg", ImGuiCol_NavWindowingDimBg),
+    Color("ModalWindowDimBg", ImGuiCol_ModalWindowDimBg),
+};
+
+// Internal Load/Dump functions
+
+static const std::string SchemaVersionLabel = "ImGuiThemeVersion";
```
WSSDude, line 309: use anonymous namespace instead of static please (again, i see this a bunch, wont fill whole review with these... please change all instances in your changes)

WSSDude: on top, better would probably be to use `constexpr auto SchemaVersionLabel  = "ImGuiThemeVersion"s` or maybe even switch to string_view for these (we use C++20, lets use it x) )

src/Theme.cpp
```diff
@@ -0,0 +1,411 @@
...
+
+static const std::string ThemeNameLabel = "ThemeName";
+static const std::string StyleSectionLabel = "style";
+static const std::string ColorsSectionLabel = "colors";
+
+static bool ThemeJsonFromStyle(json& aOutJson, const ImGuiStyle& aStyle)
+{
+    aOutJson = json{{SchemaVersionLabel, SchemaVersionValue}, {ThemeNameLabel, "<final merged theme in effect>"}, {StyleSectionLabel, {}}, {ColorsSectionLabel, {}}};
+
+    json& styleJson = aOutJson[StyleSectionLabel];
```
WSSDude, line 320: prefer auto, const auto, auto&,... instead of explicit types everywhere where the type is deducible please.

src/Theme.h
```diff
@@ -0,0 +1,12 @@
...
+// "Theme" is a configurable set of CET ImGui styling parameters
+//
+// NB: this is not a serialization of ImGuiStyle, it's a
+//     serializable representation of styling parameters.
+
+// Load ImGui style directly from JSON. Parses the format used by DumpStyleToThemeJson.
+bool LoadStyleFromThemeJson(const std::filesystem::path& aThemePath, ImGuiStyle& aOutStyle);
+
+// Dump ImGui style into a JSON string with some extra metadata. Can be saved.
+const std::string DumpStyleToThemeJson(const ImGuiStyle& aStyle);
```
WSSDude, line 12: i would just return std::string, returning const std::string would just lead to unnecessary copies (and would make more sense if you returned reference here but you do not, so std::string is better return value in this case. allows for copy ellision)

src/d3d12/D3D12_Functions.cpp
```diff
@@ -1,8 +1,9 @@
-#include <stdafx.h>
+#include "stdafx.h"
 
 #include "D3D12.h"
 #include "Options.h"
 #include "Utils.h"
+#include "Theme.h"
```
WSSDude, line 6: order includes like i once wrote somewhere

src/d3d12/D3D12_Hooks.cpp
```diff
@@ -1,6 +1,7 @@
+#include "stdafx.h"
+
 #include "CET.h"
```
WSSDude, line 3: order includes like i once wrote somewhere in this file please

WSSDude (changes requested):

src/Theme.cpp
WSSDude: little note in case you dont know - ""s -> string literal, so you can look up how it is used.
string_view has ""sv.

note that ""s also means "seconds" when chrono literals are used. dont know if we by a chance use them in some header or something, so just mentioning that there may be some conflict with these suffixes.

src/Theme.cpp
```diff
@@ -0,0 +1,411 @@
...
+    const auto cThemeJsonPath = GetAbsolutePath(aThemePath, "", false);
+
+    Log::Info("Trying to load theme from {}...", UTF16ToUTF8(cThemeJsonPath.native()));
+
+    if (!cThemeJsonPath.empty())
+    {
+        std::ifstream themeJson(cThemeJsonPath);
+        if (!themeJson)
+        {
+            Log::Info("No theme file found!");
```
WSSDude, line 392: for example here, you may just return false after logging the message. 
reduces nesting also nicely.

src/Theme.cpp
```diff
@@ -0,0 +1,411 @@
...
+
+bool LoadStyleFromThemeJson(const std::filesystem::path& aThemePath, ImGuiStyle& aOutStyle)
+{
+    bool loadedSuccessfully = false;
+
+    const auto cThemeJsonPath = GetAbsolutePath(aThemePath, "", false);
+
+    Log::Info("Trying to load theme from {}...", UTF16ToUTF8(cThemeJsonPath.native()));
+
+    if (!cThemeJsonPath.empty())
```
WSSDude, line 387: if (cThemeJsonPath.empty()) return false;

again, nesting can be reduced by something like this, just forget about loaded successfully and early-exit

maximegmd:
:'(

PR #788, opened 2023-01-31
More descriptive error log for unrecognized game version + build instructions

Murzinio:
Current message is a bit cryptic so I thought you could use something easier to understand.

### DIFF
--- a/src/Image.cpp
+++ b/src/Image.cpp
@@ -1,5 +1,8 @@
 #include <stdafx.h>
 
+#include <string>
+#include <sstream>
+
 #include "Image.h"
 
 struct PdbInfo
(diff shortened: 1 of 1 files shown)


yamashi (approved):

Murzinio:
Also added build instructions to readme.

yamashi:
The build instructions are too oriented, I don't think any of us use VScode with xmake, I expect most people contributing to this code base already have their preferences when it comes to tooling so this seems a bit excessive

Murzinio:
Sure, you can use a bunch of different editors/IDEs. This is only showing a suggestion, no one will be forced to use VS Code. 

It's not for people already contributing as they already know how to build the project but a starting point for people that want to help.

yamashi:
But we already have BUILD.md that does exactly that in way more details

Murzinio:
Sorry, missed that - I assumed it will be in the readme file.

Murzinio:
Removed the readme change.

yamashi (approved):

PR #812, opened 2023-04-16
Fix TweakDB:SetFlat

psiberx:


maximegmd (approved):
Weird but if it works x)

PR #832, opened 2023-06-20
Enable unity build also for C++

WSSDude:
We had enabled unity builds, but only for C files. This enables it for C++ also, builds should be faster.

### DIFF
--- a/src/CET.cpp
+++ b/src/CET.cpp
@@ -5,7 +5,6 @@
 
 using namespace std::chrono_literals;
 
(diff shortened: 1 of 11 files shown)


maximegmd:
It's disabled because it doesn't build ^^

WSSDude:
Yeah, cause of RED4ext though it seems... @WopsS please fix 😆

maximegmd:
Should I close this?

WSSDude:
It should be compilable still I believe with those changes, but Im unsure we want them... I would probably close this for now and leave it for another time.

PR #834, opened 2023-06-20
Make TweakDBEditor cache rebuilding async

WSSDude:
Cache rebuilding is quite slow, mainly in Debug build of CET. This causes quite a big of a lag when user opens TweakDBEditor widget for the first time. Same would happen if user hit `Refresh all` under `Advanced` tab in the widget. 

This PR delegates cache rebuilds to another thread.

### DIFF
--- a/src/overlay/widgets/TweakDBEditor.cpp
+++ b/src/overlay/widgets/TweakDBEditor.cpp
@@ -3,6 +3,7 @@
 #include "TweakDBEditor.h"
 
 #include <CET.h>
+#include <future>
 #include <Utils.h>
 
 #include <reverse/TweakDB/ResourcesList.h>
@@ -137,8 +138,10 @@ void TweakDBEditor::OnUpdate()
 
     if (!m_initialized)
     {
-        RefreshAll();
-        m_initialized = true;
+        RebuildCache();
+
+        ImGui::TextUnformatted("Rebuilding cache...");
+        return;
     }
 
     if (ImGui::BeginTabBar("TweakDBEditor-Bar"))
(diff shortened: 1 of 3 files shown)


src/overlay/widgets/TweakDBEditor.cpp
```diff
@@ -3,6 +3,7 @@
 #include "TweakDBEditor.h"
 
 #include <CET.h>
+#include <future>
```
maximegmd, line 6: not needed

src/overlay/widgets/TweakDBEditor.cpp
```diff
@@ -179,10 +182,39 @@ void TweakDBEditor::OnUpdate()
...
+        m_rebuildCacheFuture.get();
+        m_rebuildingCache = false;
+        m_initialized = true;
+
+        return;
+    }
+
+    m_rebuildingCache = true;
+
+    m_rebuildCacheFuture = std::async(
```
maximegmd, line 204: You need to set the policy to std::launch::async, otherwise you risk deferred being used as it's implementation defined

maximegmd (changes requested):

src/overlay/widgets/TweakDBEditor.cpp
WSSDude: good point, didnt think of that as async is used as default in vs

src/overlay/widgets/TweakDBEditor.cpp
WSSDude: removed

src/overlay/widgets/TweakDBEditor.cpp
WSSDude: added explicit launch type

maximegmd (approved):

PR #839, opened 2023-06-28
Revert "Feature/its all about fonts"

maximegmd:
Reverts maximegmd/CyberEngineTweaks#769

PR #840, opened 2023-06-28
Feature/its all about fonts the return

maximegmd:
Reverts maximegmd/CyberEngineTweaks#839


WSSDude:
btw, found some and resolution changes are still an issue with this.
just so there is a note here as well, Ming was contacted already about it.

PR #854, opened 2023-09-25
use updated CGameEngine-field for ClipToCenter()

Lukas0610:
Requires https://github.com/WopsS/RED4ext.SDK/pull/87 to be merged

### DIFF
--- a/src/overlay/Overlay.cpp
+++ b/src/overlay/Overlay.cpp
@@ -18,7 +18,7 @@ void Overlay::PostInitialize()
 
             auto& d3d12 = CET::Get().GetD3D12();
             d3d12.DelayedSetTrapInputInImGui(true);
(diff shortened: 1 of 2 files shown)


maximegmd:
I already fixed this in my fork of red4ext, your change to the engine class is incorrect so we should just wait for them to do a proper fix and just use my similar hack in the meantime.

maximegmd:
Yeah it is indeed at D0 but the rest of the class is probably incorrect, we need to fix the entire class

Lukas0610:
Yeah, I only patched the part for the window-data, psiberx pushed a full fix for the class just a few minutes after https://github.com/WopsS/RED4ext.SDK/pull/88/files, but he made the same changes to UnkC0 as I did, so this patch to CET should still be working as intended

PR #855, opened 2023-09-25
Feature/2.0 changes

WSSDude:


WSSDude:
crashes on exit, dont know if because of the test mod now or not
https://www.nexusmods.com/cyberpunk2077/mods/2675?tab=files&file_id=45807

WSSDude:
- crashes when launched from Steam (which launches some launcher, hitting play there causes crash...)
- crashes on exit (when launched directly from EXE)

grzeli:
@WSSDude game started from .exe file crashes, even when there are no mods.

WSSDude:
I know, found out shortly after I posted the comment about launcher crashing.
Debugger has to be attached for some reason for it to not crash.

PR #877, opened 2023-10-03
uvfs support for CET with MO2 

rfuzzo:
# The problem

When launching the game through Mod Organizer 2, CET mods don't work
That is because 
- (1) the luaVM is started too soon (before MO2s virtual file system has started)
- and (2) because MO2 overlays 3 different directories in its VFS: the base game dir, the virtualized mod dir, and the "overwrite" dir where runtime-created files are stored.

Any mod that creates files at runtime (that is ALL CET mods since the sandbox db gets created at runtime) will fail since the "overwrite" directory escapes the mod sandbox.

# This PR

This PR does two things:
1. It delays the luaVM init slightly so that CET mods get detected after UVFS did its thing
2. More problematically, it allows lua paths outside the mod sandbox under certain conditions:
  - when the game is launched through MO2 (we detect `usvfs_x64.dll`)
  - and when the lua path originates from MO2s "overwrite" directory (we detect this by comparing relative paths)

Since `GetLuaPath` does more checks now, I needed to move it out of utils into sandbox.
This also incurred pulling the sandbox into more lambdas.

### Pros: 
- It would allow running CET mods from a popular mod manager (MO2)

### Cons: 
- We give up some sandbox security under certain conditions. Other script extenders (like Bethesda Script Extenders) seem to find this acceptable.

### DIFF
--- a/src/CET.cpp
+++ b/src/CET.cpp
@@ -74,7 +74,6 @@ CET::CET()
     , m_vm(m_paths, m_bindings, m_d3d12)
     , m_overlay(m_bindings, m_options, m_persistentState, m_vm)
 {
-    m_vm.Initialize();
 }
 
 CET::~CET()
--- a/src/scripting/LuaSandbox.cpp
+++ b/src/scripting/LuaSandbox.cpp
@@ -125,6 +125,16 @@ void LuaSandbox::Initialize()
     // initialize state + environment first
     m_globals = {luaState, sol::create};
 
+    // log mo2 usage
+    m_isLaunchedThroughMO2 = GetModuleHandle(TEXT("usvfs_x64.dll")) != nullptr;
+    if (m_isLaunchedThroughMO2)
+    {
+        if (auto existingLogger = spdlog::get("scripting"))
+        {
+            existingLogger->warn("CET launched with MO2.");
+        }
+    }
+    
     // copy whitelisted things from global table
     for (const auto* cKey : s_cVMGlobalObjectsWhitelist)
     {
@@ -359,7 +369,7 @@ void LuaSandbox::InitializeIOForSandbox(Sandbox& aSandbox, const sol::state& acp
     };
     sbEnv["loadstring"] = cLoadString;
 
-    const auto cLoadFile = [cSBRootPath, cLoadString, stateView](const std::string& acPath) -> std::tuple<sol::object, sol::object>
+    const auto cLoadFile = [this, cSBRootPath, cLoadString, stateView](const std::string& acPath) -> std::tuple<sol::object, sol::object>
     {
         const auto previousCurrentPath = std::filesystem::current_path();
         current_path(cSBRootPath);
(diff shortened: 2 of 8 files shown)


src/scripting/LuaSandbox.cpp
```diff
@@ -478,7 +488,7 @@ void LuaSandbox::InitializeIOForSandbox(Sandbox& aSandbox, const sol::state& acp
         return res;
     };
 
-    sbEnv["dir"] = [cSBRootPath, stateView](const std::string& acPath) -> sol::table
+    sbEnv["dir"] = [this, cSBRootPath, stateView](const std::string& acPath) -> sol::table
```
WSSDude, line 491: passing this into capture is problematic, sandbox pointer is not guaranteedto be stable during whole runtime, that is also why we pass LuaSandbox along with index around.

WSSDude: you have this in few places

src/scripting/LuaSandbox.cpp
```diff
@@ -657,3 +667,80 @@ void LuaSandbox::CloseDBForSandbox(const Sandbox& aSandbox) const
...
+            existingLogger->info("\tresolved to: {}", _Weakly_canonical_path.string());
+        }
+#endif // CET_DEBUG
+
+        // path outside of sandbox
+        // make an exception if path outside of sandbox originates from MO2's overwrite directory
+        if (m_isLaunchedThroughMO2)
+        {
+            // simple check
+            // if (_Weakly_canonical_base.string().find("overwrite") != std::string::npos || _Weakly_canonical_path.string().find("overwrite") != std::string::npos)
```
WSSDude, line 716: why is this commented here?

src/scripting/LuaSandbox.cpp
```diff
@@ -657,3 +667,80 @@ void LuaSandbox::CloseDBForSandbox(const Sandbox& aSandbox) const
...
+        return {};
+
+    aFilePath = GetAbsolutePath(aFilePath, acRootPath, acAllowNonExisting, false);
+    if (aFilePath.empty())
+        return {};
+
+    const auto relativeFilePathToRoot = relative(aFilePath, acRootPath);
+    if (relativeFilePathToRoot.native().starts_with(L"..\\") || relativeFilePathToRoot.native().find(L"\\..\\") != std::wstring::npos)
+    {
+        // auto c = std::filesystem::current_path();
```
WSSDude, line 696: why is this commented here?

src/CET.cpp
```diff
@@ -74,7 +74,8 @@ CET::CET()
     , m_vm(m_paths, m_bindings, m_d3d12)
     , m_overlay(m_bindings, m_options, m_persistentState, m_vm)
 {
-    m_vm.Initialize();
+    // Delay our lua hook until later, to ensure that Mod Organizer's VFS is hooked up.
+    //m_vm.Initialize();
```
WSSDude, line 78: another why is this commented and here, just remove it. that is why version control like git exists.

WSSDude:
Eh, checking for .. is sketchy, they may be valid cases (like when mod from some of its subfolder does something with parent)

Dont think that is a good solution.

WSSDude:
And on top, it may appear in the middle of the path potentially and you could still escape

sth/../../yey

WSSDude:
I think this looks good apart from the few commented out stuff at least on first glance... I'd rather have a more proper look in like an hour though cause cant discern changes to the GetLuaPath as you moved + changed it in same commit.

WSSDude (approved):

PR #892, opened 2023-10-23
fix: relative file paths for uvfs

### LINKED ISSUE #890: Mod path access through Mod Organizer 2
# Bug

Some CET mods accessing files from other mods do not always work with MO2.

## Probable causes
1. PR #877 introduced a fix/workaround for MO2, which handles path resolution differently when MOs USVFS is detected: [code for overwrite path](https://github.com/maximegmd/CyberEngineTweaks/blob/7682be84d82b3cdb0adf47f833eb8f62ab24f738/src/scripting/LuaSandbox.cpp#L710-L731).  
  The issue is that MO overlays multiple mod paths, which cannot always be identified just by the path directly.

2. as seen in the [CET/CyberScript log below](#chatter) with the [related code lines](https://github.com/cyberscript77/release/blob/eefe3627eb4a42ef73a629e15d56caa91c1ac97b/bin/x64/plugins/cyber_engine_tweaks/mods/cyberscript/mod/modules/modpack.lua#L245-L286): `dir("datapack")` works, but `io.open("datapack/.../desc.json")`does not.

Game version: Cyberpunk v2.01
CET version: 1.27.1

## Suggested solution
1. Use relative paths (instead of resolving to absolute) as long as they are within the mods sub folders.
    (Or at least add handle also `mods/*/bin/x64` as a "special" MO path, although these paths can be configured differently in MO)
2. Patch also `io.open` etc. (similar to `dir`) to handle file access correctly

## Reproduction

- MO 2 beta 13 (with optional [cyberpunk plugin update v2+](https://github.com/ZashIn/modorganizer-basic_games/releases))
- CET v1.27+ installed directly into game dir

Mods (clear overwrite before each mod test):
- <a id="chatter"></a>mod [Citizen Voices and NCPD Radio Chatter](https://www.nexusmods.com/cyberpunk2077/mods/7013) + dependencies (except for CET) installed via MO
    ```
    [12236] CyberScript Assets Loaded
    [12236] No DESC FOR ..\..\..\..\..\..\..\..\Citizen Voices and NCPD Radio Chatter\bin\x64\plugins\cyber_engine_tweaks\mods\cyberscript\datapack\Citizen Voices
    [12236] Can't load ..\..\..\..\..\..\..\..\Citizen Voices and NCPD Radio Chatter\bin\x64\plugins\cyber_engine_tweaks\mods\cyberscript\datapack\Citizen Voices from the cache : 
    | jsondesc is not null : false 
    | datapack have required flag: false 
    | datapack is null : true
    ```
- [Appearance Menu Mod](https://www.nexusmods.com/cyberpunk2077/mods/790) + [V's Original ApartmentPLUS](https://www.nexusmods.com/cyberpunk2077/mods/6028) (preset):
  - with AMM in load order before preset (AMM prio < preset): preset does not show up
  - preset before AMM (preset prio > AMM): preset does show up
  - expected: preset should always be shown

All mods in `overwrite`: work

rfuzzo:
One more uvfs fix for relative file paths.

PR for @seberoths and @zashs suggestions.
~~- api addition for `dir()`: add relativePath property~~
- internal relative() resolve in luasandbox for uvfs
- restore original relativePath check for sandbox

closes #890

### DIFF
--- a/src/scripting/LuaSandbox.cpp
+++ b/src/scripting/LuaSandbox.cpp
@@ -507,7 +507,8 @@ void LuaSandbox::InitializeIOForSandbox(Sandbox& aSandbox, const sol::state& acp
         for (const auto& entry : std::filesystem::directory_iterator(path))
         {
             sol::table item(stateView, sol::create);
-            item["name"] = UTF16ToUTF8(relative(entry.path(), path).native());
+            item["relativePath"] = UTF16ToUTF8(Relative(entry.path(), path).native());
+            item["name"] = UTF16ToUTF8(entry.path().filename().native());
             item["type"] = entry.is_directory() ? "directory" : "file";
             res[index++] = item;
         }
@@ -673,6 +674,15 @@ std::filesystem::path LuaSandbox::GetLuaPath(const std::string& acFilePath, cons
     return GetLuaPath(UTF8ToUTF16(acFilePath), acRootPath, acAllowNonExisting);
 }
 
+std::filesystem::path LuaSandbox::Relative(const std::filesystem::path& acFilePath, const std::filesystem::path& acRootPath) const
+{
+    if (m_isLaunchedThroughMO2)
+    {
+        return acFilePath.lexically_relative(acRootPath);
+    }
+    return relative(acFilePath, acRootPath);
+}
+
 std::filesystem::path LuaSandbox::GetLuaPath(std::filesystem::path aFilePath, const std::filesystem::path& acRootPath, const bool acAllowNonExisting) const
 {
     assert(!aFilePath.empty());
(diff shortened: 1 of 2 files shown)


src/scripting/LuaSandbox.cpp
```diff
@@ -507,7 +507,8 @@ void LuaSandbox::InitializeIOForSandbox(Sandbox& aSandbox, const sol::state& acp
         for (const auto& entry : std::filesystem::directory_iterator(path))
         {
             sol::table item(stateView, sol::create);
-            item["name"] = UTF16ToUTF8(relative(entry.path(), path).native());
+            item["relativePath"] = UTF16ToUTF8(Relative(entry.path(), path).native());
+            item["name"] = UTF16ToUTF8(entry.path().filename().native());
```
WSSDude, line 511: this returned relative path formerly, seems a bit sus...

src/scripting/LuaSandbox.cpp
```diff
@@ -507,7 +507,8 @@ void LuaSandbox::InitializeIOForSandbox(Sandbox& aSandbox, const sol::state& acp
         for (const auto& entry : std::filesystem::directory_iterator(path))
         {
             sol::table item(stateView, sol::create);
-            item["name"] = UTF16ToUTF8(relative(entry.path(), path).native());
+            item["relativePath"] = UTF16ToUTF8(Relative(entry.path(), path).native());
```
WSSDude, line 510: do we really need this if usecase is with dir? cant we somehow just do this in the dir command if it is required for it?

src/scripting/LuaSandbox.cpp
rfuzzo: Yeah...

src/scripting/LuaSandbox.cpp
rfuzzo: I can remove it. It's not used anywhere

WSSDude (approved):

PR #900, opened 2023-10-29
Add `sol::lib::jit` to actually enable JIT ( resolves #899 )

### LINKED ISSUE #899: LuaJIT's JIT compiler is disabled by default
Ignore this if disabling the JIT compiler is the intended behavior, but the default Lua initialization does not enable the JIT compiler by default: https://github.com/maximegmd/CyberEngineTweaks/blob/91565e3a43edd6ada590bae0391d4172a39eb202/src/scripting/Scripting.cpp#L52

For the JIT to be enabled, the `sol::lib::jit` library must be opened even if the `jit` module is to be sandboxed away later:

```lua
luaVm.open_libraries(sol::lib::base, sol::lib::string, sol::lib::io, sol::lib::math, sol::lib::package, sol::lib::os, sol::lib::table, sol::lib::bit32, sol::lib::jit);
```

The LuaJIT documentation mentions this here: https://luajit.org/install.html#embed

> Make sure the `jit` library is loaded, or the JIT compiler will not be activated.

This is also mentioned in this sol2 issue: https://github.com/ThePhD/sol2/issues/61

WSSDude:
Resolves #899 .

See comment https://github.com/maximegmd/CyberEngineTweaks/issues/899#issuecomment-1784053997

### DIFF
--- a/src/CET.cpp
+++ b/src/CET.cpp
@@ -71,7 +71,7 @@ CET::CET()
     , m_bindings(m_paths, m_options)
     , m_window(&m_bindings, &m_d3d12)
     , m_d3d12(m_window, m_paths, m_options)
-    , m_vm(m_paths, m_bindings, m_d3d12)
+    , m_vm(m_paths, m_options, m_bindings, m_d3d12)
     , m_overlay(m_bindings, m_options, m_persistentState, m_vm)
 {
 }
--- a/src/Options.cpp
+++ b/src/Options.cpp
@@ -18,14 +18,9 @@ void PatchesSettings::Load(const nlohmann::json& aConfig)
 nlohmann::json PatchesSettings::Save() const
 {
     return {
-        {"disable_async_compute", AsyncCompute},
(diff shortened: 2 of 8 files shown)


WSSDude:
Just an additional note - I also try to run the test without LuaJIT in the first place, just LuaJIT about doubles the performance, but this makes it like 3x faster against just regular Lua package (Lua 5.4 version).

maximegmd:
I think we should make this an option, given that we haven't used JIT for so long we may be running into some bugs, if we can give the option to disable it, people have options until the potential issues are resolved.

WSSDude:
Good point, I'll add some setting for it. Will we enable it by default?

maximegmd:
I think we should enable it by default and explain that people can disable it if they encounter new errors

WSSDude:
Added it as an option, defaulted to true 👍

maximegmd (approved):

Issue #888, opened 2023-10-17
Incorrect int handling in Enum class


googleben:
# Bug in `Enum` class

Operating system: Windows
Game version: 2.01
CET version: 1.27.1
GPU: RTX 4080

## The Problem

In the CET Enum class, there's some missed casts that result in enums not working correctly in Lua. I'll use `EDeviceStatus` for example. `EDeviceStatus` is a 32-bit enum, as defined by `CEnum.actualSize` which is 4 in this case. However, all the values for enums are stored as 64-bit `int64_t`s in the field `CEnum.valueList` (and `CEnum.aliasValueList`), type `DynArray<int64_t>` (see [here](https://github.com/WopsS/RED4ext.SDK/blob/9df79fd6b51ab262ea17e070a4bf1b1b034a34e9/include/RED4ext/RTTITypes.hpp#L293)) which means we need to do some type casting pretty much any time we interact with those fields. Looking at [the code for the CET `Enum` class](https://github.com/maximegmd/CyberEngineTweaks/blob/7682be84d82b3cdb0adf47f833eb8f62ab24f738/src/reverse/Enum.cpp), we can see this at least seems to be done correctly in the `get` and `set` methods, but isn't done in any other method.

Take for example `EDeviceStatus.UNPOWERED`, which is -1 or 4294967295 if unsigned. You can check that this is the correct value by finding an unpowered door in the game and running this Lua code:

```Lua
print(Game.GetTargetingSystem():GetLookAtObject(Game.GetPlayer(), false, false):GetDevicePS():GetDeviceState())
```

[...]

PR #902, opened 2023-11-09
Fix integer sizing in Enum

googleben:
Fixes https://github.com/maximegmd/CyberEngineTweaks/issues/888.

Pretty straightforward changes - mainly just converted `Enum.m_value` to a union and wrote conversions where needed. Important bits:

- Two constructors still only take `int32_t` which theoretically isn't wide enough, but they're currently only called using numbers from Lua which aren't precise enough either. Could be fixed, but a cursory scan didn't turn up anything currently in the game that's large enough to need more than an int32, so I didn't worry about it
- I went ahead and switched to all signed numbers. I haven't seen anything that seems unsigned, but there's quite a lot of obviously negative numbers, so this seems reasonable to me
- I modified the `EnumInt` function exposed to Lua to return a correctly sized int. This isn't strictly necessary and it could just be an Int64 if that's preferred
- For testing I created enums for every size that each had values of 0, 1, 2, -1, -2, and the max and min values for the type. Everything seemed to work correctly, except of course that Lua can't represent max int64_t with its native numbers
- I made sure all the `Enum` constructors work correctly from Lua

Let me know if there's any changes I need to make or other information you need!

### DIFF
--- a/src/reverse/Converter.h
+++ b/src/reverse/Converter.h
@@ -156,7 +156,7 @@ struct EnumConverter : LuaRED<Enum, "Enum">
         else if (aObject.get_type() == sol::type::number) // Enum from number cast
         {
             auto* enumType = static_cast<RED4ext::CEnum*>(apType->type);
-            const Enum en(enumType, aObject.as<uint32_t>());
+            const Enum en(enumType, aObject.as<int32_t>());
             en.Set(*apType);
         }
         else if (aObject.get_type() == sol::type::string) // Enum from string cast
--- a/src/reverse/Enum.cpp
+++ b/src/reverse/Enum.cpp
@@ -17,7 +17,7 @@ Enum::Enum(const std::string& acTypeName, const std::string& acValue)
     }
 }
 
-Enum::Enum(const std::string& acTypeName, uint32_t aValue)
+Enum::Enum(const std::string& acTypeName, int32_t aValue)
 {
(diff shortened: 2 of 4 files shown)


WSSDude:
LuaJIT provides ability to do this, we have some converter for 64-bit values. They have to be suffixed, we have it hacked a bit that it creates specific lua string and executes it to get correct type. There is probably better way but this is what we have and do.

WSSDude:
https://github.com/maximegmd/CyberEngineTweaks/blob/0544f6fc49862e8707ea8b014a4304e0a8bafc5f/src/reverse/LuaRED.h#L12

WSSDude:
If it doesnt work for some reason, it should be looked at, we need to be able to work with 64 bit numbers from Lua in some way.

If on the Lua side you tried to just put some big number, try to add that ll/ull suffix.

If this fails getting to Lua properly, I would have a look if it doesn't fail to convert properly, but I would guess this is called somewhere down the line and it should work, that it is a suffix issue on Lua side (you didn't include it so it didnt know what to do).

WSSDude:
I did not have a look at the changes btw, would like to do so later... Think this needs thorough look, need to make sure you do static conversions correctly (that may be the biggest issue with such solution). Just skimmed through it, saw some safe sets/gets...

Just reacting to what you wrote now more or less, dont take it as a review :P

googleben:
Appending `ll` did the trick - didn't realize I could do that lol. I think that also means I don't have any more concerns about retyping those 2 constructors to use `int64_t`, so two birds with one stone. I'll double check that and retest it when I get a chance.

PR #910, opened 2023-12-25
Fix "Disable VSync" settings description

PotatoOfDoom:
I fixed the description of the Disable V-sync setting because it was seemingly copy-pasted to the wrong Fix Minimap Flicker setting. I modified that settings description too but am not quite sure what to write in there because I never actually experienced the issue myself.


WSSDude:
Flicker fix is meant to be used with some mods modifying minimap which otherwise caused flickering, unsure which one it was now nor if it is relevant still tbh. :suspect:

WSSDude (approved):

PR #912, opened 2024-01-07
Expose mod input handlers for better integration between mods

Supremist:
Allow mods to:
 - Get name list of other installed mods
 - Get info about input handlers, registered for other mods
 - Call input handlers from lua, as if user presses an input key
 - Split responsibilities: mods will delegate input management to other specific mods

Current CET implementation of Input & Hotkey functionality provides a good default for mods. But it ignores the game context in which the key was pressed. That means, if the key is bound in CET then it must be unbound in game. That is ok, but with enough mods users will run out of keys they can comfortably press.

My proposal is to allow input mods to monitor in game context and invoke other mods key bindings depending on context. So a single key can do different action, when the player is in a vehicle or while sprinting, for example. 
Of course each keybind should be EITHER bound in CET OR managed by separate input mod.


maximegmd (approved):

src/scripting/Scripting.cpp
```diff
@@ -599,11 +609,45 @@ void Scripting::TriggerOnOverlayClose() const
...
+    {
+        sol::table info(lockedState.Get(), sol::create);
+        info["slug"] = bind.ID;
+        info["label"] = bind.DisplayName;
+        info["callback"] = bind.VmHandler;
+        if (std::holds_alternative<std::string>(bind.Description))
+        {
+            info["description"] = std::get<std::string>(bind.Description);
+        }
+        info["bind"] = bindings.GetBindString({acName, bind.ID});
```
WSSDude, line 645: You cant really get valuable data sometimes from this, CET doesnt handle properly some extended keys, writting them as Unknown into string. + Im a bit unsure what would be the use-case for this bit, or anything here in general. Some sample mod/mods showing how this should work in practics would be nice.

src/scripting/Scripting.cpp
```diff
@@ -599,11 +609,45 @@ void Scripting::TriggerOnOverlayClose() const
...
+    if (!modBinds)
+        return sol::nil;
+
+    auto lockedState = m_lua.Lock();
+    const auto& bindings = m_store.GetBindings();
+    sol::table res(lockedState.Get(), sol::create);
+    for (const auto& bind : *modBinds)
+    {
+        sol::table info(lockedState.Get(), sol::create);
+        info["slug"] = bind.ID;
```
WSSDude, line 638: Why calm it slug? Just call it id.

src/scripting/Scripting.cpp
```diff
@@ -599,11 +609,45 @@ void Scripting::TriggerOnOverlayClose() const
...
+        info["label"] = bind.DisplayName;
```
WSSDude, line 639: Also would not unnecessarily call this something else, there was some plan for localization, it should be displayName, it is not really meant to be a label.

src/scripting/Scripting.cpp
```diff
@@ -599,11 +609,45 @@ void Scripting::TriggerOnOverlayClose() const
...
+        res.add(name);
+    return res;
+}
+
 sol::object Scripting::GetMod(const std::string& acName) const
 {
     return m_store.GetMod(acName);
 }
 
+sol::object Scripting::GetModInputHandlers(const std::string& acName) const
```
WSSDude, line 626: Bindings can be hotkeys (up to 4 key combo, they activate always on key up) or inputs (single key, is called on key down/up). I dont see you use or work with this info. Also, I suppose this should return whether it is one or the other.

src/scripting/ScriptStore.h
```diff
@@ -23,6 +24,7 @@ struct ScriptStore
     void TriggerOnOverlayOpen() const;
     void TriggerOnOverlayClose() const;
 
+    [[nodiscard]] TiltedPhoques::Vector<std::string> GetModNames() const;
```
WSSDude, line 27: I'd probably make this into something like GetMods rather, which would return some mod info (would make it some table, for now, I suppose only name would be inside of it) and object normally returned by GetMod. Think it would be better personally, and we can extend it with proper mod info when we get to it.

src/scripting/ScriptContext.cpp
```diff
@@ -115,7 +115,7 @@ ScriptContext::ScriptContext(LuaSandbox& aLuaSandbox, const std::filesystem::pat
         return std::get<std::string>(acDescription);
     };
 
-    auto registerBinding = [this, &wrapHandler, &wrapDescription](
+    auto registerBinding = [this, &wrapHandler, &wrapDescription, &lockedState](
```
WSSDude, line 118: you can just copy lockedState, it is simple pointer

WSSDude (changes requested):

WSSDude:
I need to take more closer look at this, these are just some quick remarks.

src/scripting/Scripting.cpp
WSSDude: Actually, if you were to do the GetMod change, I suppose we could also return handlers directly in it...

Also thinking now if it wouldnt make sense to introducd breaking change and make GetMod return single item from what GetMods would return, further simplifying the usage of this (table with mod info, bindings and return value from init.lua of the mod)

WSSDude:
I added an extra comment and edited some for better clarity (you won't see changes from email, so check them in the PR please)

src/scripting/ScriptContext.cpp
```diff
@@ -145,8 +145,18 @@ ScriptContext::ScriptContext(LuaSandbox& aLuaSandbox, const std::filesystem::pat
...
+        sol::function vmHandler = aCallback;
+        if (acIsHotkey)
+        {
+            vmHandler = MakeSolFunction(lockedState, [aCallback](bool isPressed) {
+                if (isPressed)
+                {
+                    aCallback();
+                }
+            });
+        }
```
Supremist, line 157: Here is the code, that handles hotkeys. Personaly, I would rather expose single interface for both hotkey and input callback - `void callback(bool isPressed)`. But if you see an issue with that, I will make two separate fields `isHotkey` and `callback`. What do you prefer?

src/scripting/Scripting.cpp
Supremist: Ok, `GetBindString` is not really what I need. I guess I'll change it to
```info["isBound"] = bindings.IsBound({acName, bind.ID});```
to be able to just check if input is bound in CET or not and is it safe to call from my mod.
I am still experimenting with input to be honest, but I think I can show some quick and dirty mod, that shows what I want to do with this functionality.

File: src/scripting/Scripting.cpp (hunk 3 above)
Comment (Supremist) [reply src/scripting/Scripting.cpp:639]: You right, I'll rename it. I was just looking at names in wiki:
> -- registerInput()
> --
> -- @param  string    slug      The internal slug (must be unique in your mod scope)
> -- @param  string    label     The label displayed in CET Bindings
> -- @param  function  callback  The callback function

File: src/scripting/ScriptContext.cpp (hunk 6 above)
Comment (Supremist) [reply src/scripting/ScriptContext.cpp:118]: it is a shared pointer with a scoped lock, that is why I can't copy it. I can copy just a pointer or capture whole luaSandbox and lock inside. But I rather leave it as it is.

File: src/scripting/ScriptContext.cpp (hunk 7 above)
Comment (WSSDude) [reply src/scripting/ScriptContext.cpp:157]: Hotkeys react on key up as mentioned before, otherwise they would trigger in wrong moment (imagine binds like ctrl + shift and ctrl + shift + a, it would register both if it was on key down or completely ignore the longer one)

File: src/scripting/ScriptContext.cpp (hunk 6 above)
Comment (WSSDude) [reply src/scripting/ScriptContext.cpp:118]: ah you right had not realised that... state in sol3 is but this wraps it. k, leave it as is.

File: src/scripting/Scripting.cpp (hunk 3 above)
Comment (WSSDude) [reply src/scripting/Scripting.cpp]: wiki is community driven, we had not wrote that... ill notify somebody or make an edit later.

File: src/scripting/Scripting.cpp (hunk 1 above)
Comment (WSSDude) [reply src/scripting/Scripting.cpp]: would be nice to see some expected usage, ideally with fixes. maybe sth can be done nicer/better, main point why i wanted some sample :)

File: src/scripting/Scripting.cpp (hunk 4 above)
Comment (WSSDude) [reply src/scripting/Scripting.cpp]: Great, thanks! Now to just move the inner loop into ScriptContext and make GetMod also use it and I think this will be as clean as it gets.

File: src/scripting/ScriptContext.cpp (hunk 7 above)
Comment (Supremist) [reply src/scripting/ScriptContext.cpp:157]: Yes, that is how CET handles hotkeys internally.
Let's imagine communication between two mods: 
In "clientMod":
```registerHotkey("actionId", "Custom Action", function() 
...
end)```

In "InputMod" when certain game input action event is detected:
```GetMods()["clientMod"].inputHandlers["actionId"].callback(true) -- pressed
```

In "InputMod" when certain game input action finished:
```GetMods()["clientMod"].inputHandlers["actionId"].callback(false) -- released
```
I want the "InputMod" to not care about what callback it is, hotkey or not.
So I am not changing what CET is doing with hotkeys, just adding new api for mods like "InputMod".
Alternative would be to do it like this:

```
local handler = GetMods()["clientMod"].inputHandlers["actionId"]
if handler.isHotkey then
    handler.callback()
else
   handler.callback(true)
end
```

WSSDude: Yeah like that seems fine. Only issue I had with having it in same handler is that hotkey should trigger when key is up (so when false is passed into the function, you react when key is down so when true is passed in)

Think it would just confuse things tbh, when real behavior is one but this is different.

src/scripting/Scripting.cpp
Supremist: I created a simple [mod](https://github.com/Supremist/BetterControl) for this. It enables to use Q end E for lean_left/lean_right while aiming. And when player is not aiming - the buttons do default actions (so Q does quick melee)

src/scripting/ScriptContext.cpp
Supremist: Yes, your point makes sense to me. You are thinking about how it already works in CET. So it would be confusing if hotkeys would trigger on press if they are already triggering on release.

But I am thinking about what interface we are giving to the mods. So my logic is:
 - when you want activation: call `callback(true)`
 - when you want deactivation (if that is even supported): call `callback(false)`
So in that sense, hotkeys do not support deactivation. That would be the only difference from inputs.

After your proposal 
> (so when false is passed into the function, you react when key is down so when true is passed in)
Activation would look like:
```
handler.callback(not handler.isHotkey)
```
If you insist on changing that, I would rather remove the if statement, and leave just 
```
sol::function vmHandler = aCallback;
```

Supremist:
Also, I have to emphasize, that now this PR introduce a braking change in the `GetMod` function. Just reiterating, to be clear.

src/scripting/ScriptContext.cpp
```diff
@@ -294,6 +304,38 @@ sol::object ScriptContext::GetRootObject() const
     return m_object;
 }
 
+sol::object ScriptContext::GetInputHandlers() const
+{
+    auto lockedState = m_sandbox.GetLockedState();
+    const auto& bindings = m_sandbox.GetBindings();
```
WSSDude, line 310: I'd probably just use CET::Get().GetBindings() here, wouldnt necessarily make another GetBindings. If something needs them, they should be passed down (if it was done the right way). We have few places like this already, there should be some cleanup to clean majority of CET::Get() tbh, is fugly... So rather use that, so it is easier to find when the time comes. No need to change this now... Just dont introduce another getter for same thing please.

src/scripting/ScriptContext.cpp
WSSDude: Well, I would because I still think that this should basically emulate the lower level of things if it should be exposed. Lower level behaves like it behaves. Trying to flip the behaviour wouldn't lead to any good in the long run imho. 

I suppose as you now expose the isHotkey, just use the calback directly. Anyone can resolve it on the Lua side how they want afterwards.

src/scripting/Scripting.cpp
WSSDude: Thanks, will try it with the PR when I get home to see how things work/look in practise 👍

WSSDude (approved):

Supremist:
I’m sorry to see my PR gets reverted. Can I help to improve it?
Possible solutions:
 1. Introduce new function and do not introduce a breaking change in the existing function (as it was implemented before the review)
 2. Merge to different branch with breaking changes and release it later after deprecating the current branch (I don’t really like it, as it is too slow and forcing for other mod authors)
 3. Introduce an API version and bind mod to different APIs depending on what version the mod requests. (I like it the most, but it is more work)

WSSDude:
Apparently, more mods used GetMod than I expected... So new functions which do that would be better probably, GetModContext+GetModContexts probably.

GetMod could be wired to GetModContext and return rootObject to not break former bindings.

WSSDude:
This was only reason this was reverted, no other really... If it was some separate function in the end, it would be fine.

PR #916, opened 2024-01-20
Add BasicTypes overload operators

poirierlouis:
Add overload operators for `Vector3` and `Vector4`.

- Includes +, -, *, /, < and <= operations.
- Add `ThrowLuaError` in `LuaVM` to throw when dividing by zero.
- Formats Lua usertypes in `Scripting` for easier maintenance.

2024-01-20_CET_basictypes-overload-operators

### DIFF
--- a/src/reverse/BasicTypes.cpp
+++ b/src/reverse/BasicTypes.cpp
@@ -16,6 +16,41 @@ bool Vector3::operator==(const Vector3& acRhs) const noexcept
     return x == acRhs.x && y == acRhs.y && z == acRhs.z;
 }
 
+bool Vector3::operator<(const Vector3& acRhs) const noexcept
+{
+    return x < acRhs.x && y < acRhs.y && z < acRhs.z;
+}
+
+bool Vector3::operator<=(const Vector3& acRhs) const noexcept
+{
+    return x <= acRhs.x && y <= acRhs.y && z <= acRhs.z;
+}
+
+Vector3 Vector3::operator+(const Vector3& acRhs) const noexcept
+{
+    return Vector3(x + acRhs.x, y + acRhs.y, z + acRhs.z);
+}
+
+Vector3 Vector3::operator-(const Vector3& acRhs) const noexcept
+{
+    return Vector3(x - acRhs.x, y - acRhs.y, z - acRhs.z);
+}
+
+Vector3 Vector3::operator*(const Vector3& acRhs) const noexcept
+{
+    return Vector3(x * acRhs.x, y * acRhs.y, z * acRhs.z);
+}
+
+Vector3 Vector3::operator/(const Vector3& acRhs) const noexcept
+{
+    if (acRhs.x == 0 || acRhs.y == 0 || acRhs.z == 0)
+    {
+        CET::Get().GetVM().ThrowLuaError("Divide by zero exception");
+        return acRhs;
+    }
+    return Vector3(x / acRhs.x, y / acRhs.y, z / acRhs.z);
+}
+
 std::string Vector4::ToString() const noexcept
 {
     return fmt::format("ToVector4{{ x = {}, y = {}, z = {}, w = {} }}", x, y, z, w);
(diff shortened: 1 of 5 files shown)


WSSDude (dismissed):
Game already has these operators defined, and I bet/hope they are done in a bit more performant way than what you do here... Unsure if this should be accepted tbh, if I would much rather somehow rewire game functions already handling these as the operators. Am not a big fan of this PR overall to say the least in its current form.

WSSDude:
I also think that the ovetloads should be added to RED4ext.SDK and used from there, wouldnt add more specializations into CET when the base is that, just makes things more cumbersome to maintain.

maximegmd:
Looks good to me, I don't see how this could be made more performant, the compiler will do the SIMD optimization and calling the game function has a cost.

WSSDude:
Unsure if this will get much SIMD optimization and unsure what is the cost of building the new object... Considering this was meant to shadow existing game functions anyway, I at least think it would be better to reuse them. I'll pull my review if you want to merge it but dont find this particularly good personally, reimplementing things... 

Also think this should be part of RED4ext but that was another point, not performance-related.

But

WSSDude:
Didnt mean to send it... At least I would change the comparison operators to take into account some form of epsilon if nothing else. Same for other comparisons, is not the best to compare floating points directly like this.

WSSDude:
Nah, ended up editing my comments anyway to add few points... Too late to apologise for multiple messages...

Dismissed my review, do what you must based on my comments I guess.

poirierlouis:
About performance, I might write a tiny benchmark to compare time of execution between native methods and Lua implementation. See if there is a significant difference and, I guess, avoid any assumption on this matter. Maybe even detect different results...

About epsilon comparison, I was wondering if it should be used. Currently, it is not used with operator `==`. But as you point out, it might be within native implementation. Hence, it might be better to rely on native implementation only.

About RED4ext.SDK implementation, if it can be directly implemented there and cascade to Lua without much effort, well yes. The though didn't cross my mind when talking about this missing "feature" in Lua. I'm wondering, currently files for [Vector3](https://github.com/WopsS/RED4ext.SDK/blob/master/include/RED4ext/Scripting/Natives/Generated/Vector3.hpp) and [Vector4](https://github.com/WopsS/RED4ext.SDK/blob/master/include/RED4ext/Scripting/Natives/Generated/Vector4.hpp) are generated. Would it be possible to write operators there without messing the code when generating them?

poirierlouis:
Here a [tiny-benchmark.txt](https://github.com/maximegmd/CyberEngineTweaks/files/14001205/tiny-benchmark.txt) you can rename with `.lua` extension and try yourself.

First test is most likely too simple regarding floating point precision... However difference between overload operators and call to native functions speak for itself IMO (loop of 100 000 iterations). Of course it is even worse when you get and call global function without storing a reference before running computation. So I didn't bother to let it in attachment. Elapsed time here is in seconds:

2024-01-21_CET_basictypes-overload-operators-tiny-benchmark

Feel free to improve this benchmark if it is too simple to be representative or point out any mistakes.

poirierlouis:
Here a [tiny-benchmark.txt](https://github.com/maximegmd/CyberEngineTweaks/files/14001300/tiny-benchmark.txt) you can rename with `.lua` extension and try yourself.

First test is most likely too simple regarding floating point precision... However difference between overload operators and call to native functions speak for itself IMO (loop of 100 000 iterations). Of course it is even worse when you get and call global function without storing a reference before running computation. So I didn't bother to let it in attachment. Elapsed time here is in seconds:

2024-01-21_CET_basictypes-overload-operators-tiny-benchmark

Feel free to improve this benchmark if it is too simple to be representative or point out any mistakes.

poirierlouis:
See [discussion](https://github.com/WopsS/RED4ext.SDK/discussions/104) on RED4ext.SDK.

poirierlouis:
I'll update this PR after this is done https://github.com/WopsS/RED4ext.SDK/pull/140.

maximegmd:
Can I close this or is there a plan to update this @poirierlouis ?

poirierlouis:
I can update this weekend and include methods added in SDK too, if that's ok?

maximegmd:
Sure that works :)

PR #920, opened 2024-01-31
Revert "Expose mod input handlers for better integration between mods"

maximegmd:
Reverts maximegmd/CyberEngineTweaks#912


maximegmd:
PR broke GetMod API, I suggest adding another function to keep the GetMod API stable.

Issue #957, opened 2024-07-31
Overlay mod bindings and access to mouse/keyboard


WSSDude:
We disallowed access to keyboard and mouse functions from our fork of sol_ImGui.

I believe, considering ImGui is completely inaccessible from outside of onDraw now, we could reenable keyboard and mouse access to it.

There were few valid requests for such access, like having an ability to have some copy-paste functionality or custom shortcuts while overlay is active.

I would propose reenabling the disabled functions (believe they are just left commented out inside sol_ImGui) and potentially tweaking VKBindings to allow for modbinds separate from in-game ones which would get triggered while overlay to allow for overlay-specific shortcuts.

This could be split into two separate tasks/PRs.

I believe sol_ImGui alone should cover most use-cases, but would need to get properly tested and implemented.

Making the issue so someone can pick it up when they want to. This enhancment would be something that should now be safe and I believe not that hard to implement. Marking it as an enhamcement and good first issue, as I believe it should be quite a small change, at least for the sol_ImGui part.

WSSDude:
If anyone is willing to pick this up, can provide help as this is something I majorily worked on for CET in the past, know the systems and expected behaviors.

Feel free to contact me.

WSSDude:
Interesting commit for this task:
https://github.com/maximegmd/CyberEngineTweaks/commit/4f23722a5fc68a65c6aa5ef2ac12e0846d9c1164

Everything that was removed by this commit that was under `SOL_IMGUI_ENABLE_INPUT_FUNCTIONS` should get reverted and retested it is
* only accessible from onDraw
* works as expected

I suspect 0 to none issues from such partial revert and removal of the define, just dont revert whole commit...

PR #958, opened 2024-08-03
Added back ImGui MKB Input functions, and added new "Overlay Hotkey" category (Issue #957)

justarandomguyintheinternet:
- Implemented and made the changes as described in #957
- Reverted the changes from 4f23722, bringing back ImGui MKB Input functions
- Made changes to allow for a sub-type of Hotkeys: Overlay Hotkeys
    - These only get activated while the CET overlay is active
- They can be registered from LUA via `registerOverlayHotkey(id, name, callback)`


WSSDude:
Will need to have proper look tomorrow but I would want some changes definitely. Overall doesnt seem bad though and seems to cover #957 quite nicely :+1:

src/scripting/ScriptContext.cpp
```diff
@@ -117,7 +117,7 @@ ScriptContext::ScriptContext(LuaSandbox& aLuaSandbox, const std::filesystem::pat
 
     auto registerBinding = [this, &wrapHandler, &wrapDescription](
                                const std::string& acID, const std::string& acDisplayName, const std::variant<std::string, sol::function>& acDescription, sol::function aCallback,
-                               const bool acIsHotkey)
+                               const bool acIsHotkey, const bool acIsOverlayHotkey)
```
WSSDude, line 120: I would rather have some enum with entries like 
* Input
* Hotkey
* OverlayHotkey
Instead of two weird bools when we have 3 options atm (even with two was a bit hacky choice by me back than...)

src/scripting/ScriptContext.cpp
```diff
@@ -146,22 +146,29 @@ ScriptContext::ScriptContext(LuaSandbox& aLuaSandbox, const std::filesystem::pat
             return;
         }
 
-        m_vkBinds.emplace_back(acID, acDisplayName, wrapDescription(acDescription), wrapHandler(aCallback, acIsHotkey));
+        m_vkBinds.emplace_back(acID, acDisplayName, wrapDescription(acDescription), wrapHandler(aCallback, acIsHotkey), acIsOverlayHotkey);
```
WSSDude, line 149: wrapHandler should be accepting the enum and make a proper wrapper, there should be another type added into the variant that inherits from normal hotkey without any special extensions

src/VKBindings.h
```diff
@@ -19,6 +19,7 @@ struct VKBind
     std::string DisplayName{};
     std::variant<std::string, std::function<void()>> Description{};
     std::variant<std::function<TVKBindHotkeyCallback>, std::function<TVKBindInputCallback>> Handler{};
```
WSSDude, line 21: the variant should copy the enum, having Input, Hotkey and OverlayHotkey options, instead of boolean differentiating between hotkey and overlayhotkey

i know the order is a bit different now but swap it when you will change it

WSSDude (changes requested):
just some quick review but i want a more proper look as said, this is really surface level

WSSDude:
Do overlay keys not block ImGui input? Was this tested fully? I have a feel it may forbid from typing some letters but yeah, that is one of the things I need to properly "review"...

And another thing - did you verify that input functions of ImGui are not accessible outside of onDraw right? Think we move whole ImGui table outside of onDraw but better be sure...

Another thing I need to properly look at...

Guess other things will get resolved when you change it according to comments but will write them down tomorrow if they wont change in the meantime.

src/scripting/ScriptContext.cpp
WSSDude: looking at the definition of hotkeys, they are just std::function<type>...

probably for this to work correctly, you would want to change the usings into some actual structs holding an std::function first

WSSDude:
One more note following Discord messages - ImGui keyboard functions may not work with other layout than QWERTY I realised, would need to be checked...

If it doesnt function with other layout, would mark it as limitation for now as Im unsure now if it works with other layouts without changes in imgui_impl/win32.cpp (which this shouldnt touch ideally... that should be left for some other PR if it is an issue...)

log.txt
WSSDude, line 1: this shouldnt be here ger rid of it

src/sol_imgui/sol_imgui.h
WSSDude, line 1: also modify readme file please

src/VKBindings.h
WSSDude: also misses IsOverlayHotkey function and corresponding Lua callback (believe there was callback for it, will need to check)

i would also add probable IsValid, which would be true if variant is not std::monostate when you update it

WSSDude (changes requested):

src/sol_imgui/sol_imgui.h
WSSDude: could probably split the sol_ImGui to separate PR, both header and readme, as originally suggested in #957

justarandomguyintheinternet:
> Do overlay keys not block ImGui input? Was this tested fully? I have a feel it may forbid from typing some letters but yeah, that is one of the things I need to properly "review"... Like if I set some hotkey like A+B, will I still be able to type A+B in ImGui input fields?
> 
> And another thing - did you verify that input functions of ImGui are not accessible outside of onDraw right? Think we move whole ImGui table outside of onDraw but better be sure...
> 
> Another thing I need to properly look at...
> 
> Guess other things will get resolved when you change it according to comments but will write them down tomorrow if they wont change in the meantime.

- Yes, I did test that. Keys that are used in an overlay hotkey still get through, unless the full hotkey is pressed, in which case the keys get intercepted (Should be expected)
- Also tested that, calling any of the ImGui input functions outside `onDraw` throws the expected `Tried to call ImGui from invalid event!` error

src/sol_imgui/sol_imgui.h
justarandomguyintheinternet: Added back input functions to readme: 5d282d17a81e9428beff80d4dec24263454f86db

log.txt
justarandomguyintheinternet: Removed in 2d4bb2a89410d955ef24722dcfea526dc3933f32

manavortex (approved):
LGTM

PR #964, opened 2024-09-13
Small fixes for Patch 2.13

alphanin9:
- Updated RED4ext.SDK
- Updated AddressHashes::CPhotoMode_SetRecordID

- Note: CScript_RegisterMemberFunction seems to not be used by anything

Seems to work by my untrained eye, I can run around using CET mods and spawn NPCs

### DIFF
--- a/src/Image.h
+++ b/src/Image.h
@@ -8,7 +8,7 @@ struct Image
 {
     void Initialize();
 
-    static std::tuple<uint32_t, uint16_t> GetSupportedVersion() noexcept { return std::make_tuple(2, 1201); }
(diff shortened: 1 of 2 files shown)


dragonzkiller (changes requested):
In `src/Image.h` you need to update `GetSupportedVersion`

```cpp
static std::tuple<uint32_t, uint16_t> GetSupportedVersion() noexcept { return std::make_tuple(2, 13); }
```

Everything else seems to look good to me. Odd that they removed the other hash and then reverted back to the previous one.

alphanin9:
Done

maximegmd (approved):

PR #970, opened 2024-10-25
Update ImGui to 1.91.1

dragonzkiller:
Updates ImGui from 1.88 to 1.91.1

sol2 lua bindings were only updated to ensure build and to rename and remove bad enums--**NOT** to add any new features (I'll need help with that). The removed *functions* are rewritten in the sol2 binding to be their current equivalent, but the result could be slightly different. We should consider ways to mention these functions are now deprecated for CET.


src/d3d12/D3D12.h
```diff
@@ -3,6 +3,8 @@
 #include "common/D3D12Downlevel.h"
 #include "window/Window.h"
 
+extern IMGUI_IMPL_API LRESULT ImGui_ImplWin32_WndProcHandler(HWND hWnd, UINT msg, WPARAM wParam, LPARAM lParam);
```
WSSDude, line 6: this should not be required if you update it right, unsure why you need it

src/d3d12/D3D12_Functions.cpp
```diff
@@ -466,7 +464,7 @@ bool D3D12::InitializeImGui(size_t aBuffersCounts)
 
     ReloadFonts();
 
-    if (!ImGui_ImplDX12_CreateDeviceObjects(m_pCommandQueue.Get()))
+    if (!ImGui_ImplDX12_CreateDeviceObjects())
```
WSSDude, line 467: as said on discord, we need to pass in command queue...

src/d3d12/D3D12_Functions.cpp
```diff
@@ -484,7 +482,7 @@ void D3D12::PrepareUpdate()
 
     std::lock_guard _(m_imguiLock);
 
-    ImGui_ImplWin32_NewFrame(m_outSize);
+    ImGui_ImplWin32_NewFrame();
```
WSSDude, line 485: same for size, otherwise breaks with DPI scaling on anything other than 100%

WSSDude (changes requested):

src/d3d12/D3D12_Functions.cpp
dragonzkiller: Newer versions of ImGui have some level of DPI awareness in the Win32 impl. I can add this back easily though if we need to pass it back. I have DPI scaling on my computer and it seems to work fine without this. Similar to how it was looking in the current version.

src/d3d12/D3D12.h
dragonzkiller: This is because of how the new Win32 impl is written. It's wrapped in `#IF 0`, and I just put it here since it's where it's used. See: https://github.com/maximegmd/CyberEngineTweaks/pull/970/files/dea67f4af117072d46e8b799c4a1770f9c550726#diff-e3de0213e18b0bd55bf53f2687e616d36dd853e460ba511e634cae6c4671e981L28

src/d3d12/D3D12_Functions.cpp
dragonzkiller: I'll add this back. What is this doing?

src/d3d12/D3D12_Functions.cpp
WSSDude: as long as ImGui seem to report correct coords I believe this should be fine then

src/d3d12/D3D12_Functions.cpp
WSSDude: helps with hdr, crashing when alt+tabing etc (framebuffer resizes) a bit (not fully though), better reshade compatibility as we do not try to make bastardized command queue depending on who overrides dx functions, better windows 7 compat as either downlevel or normal command queue is used

a lot simply put tiny things together this was required for

xmake.lua
```diff
@@ -64,7 +64,7 @@ target("RED4ext.SDK")
     on_install(function() end)
 
 target("cyber_engine_tweaks")
-    add_defines("WIN32_LEAN_AND_MEAN", "NOMINMAX", "WINVER=0x0601", "SOL_ALL_SAFETIES_ON", "SOL_LUAJIT=1", "SOL_EXCEPTIONS_SAFE_PROPAGATION", "SPDLOG_WCHAR_TO_UTF8_SUPPORT", "SPDLOG_WCHAR_FILENAMES", "SPDLOG_WCHAR_SUPPORT", "IMGUI_USER_CONFIG=\""..imguiUserConfig.."\"") -- WINVER=0x0601 == Windows 7xmake
+    add_defines("WIN32_LEAN_AND_MEAN", "NOMINMAX", "WINVER=0x0601", "SOL_ALL_SAFETIES_ON", "SOL_LUAJIT=1", "SOL_EXCEPTIONS_SAFE_PROPAGATION", "SPDLOG_WCHAR_TO_UTF8_SUPPORT", "SPDLOG_WCHAR_FILENAMES", "SPDLOG_WCHAR_SUPPORT", "IMGUI_USER_CONFIG=\""..imguiUserConfig.."\", IMGUI_IMPL_WIN32_DISABLE_GAMEPAD") -- WINVER=0x0601 == Windows 7xmake
```
WSSDude, line 67: i would just define this in the user config, it should have exactly the same effect

src/sol_imgui/sol_imgui.h
```diff
@@ -58,6 +60,18 @@ inline bool BeginChild(const std::string& name, float sizeX, float sizeY, bool b
...
     return ImGui::BeginChild(name.c_str(), {sizeX, sizeY}, border, flags);
 }
+
+// new
+inline bool BeginChild(const std::string& name, float sizeX, float sizeY, int child_flags)
+{
+    return ImGui::BeginChild(name.c_str(), {sizeX, sizeY}, static_cast<ImGuiChildFlags>(child_flags));
+}
+
+inline bool BeginChild(const std::string& name, float sizeX, float sizeY, int child_flags, int window_flags)
```
WSSDude, line 70: ImVec2 does not function with sol? i thought i fixed that, would be nice probably to have them exposed also if they work as possible overload.

if not, dont bother, is fine, just an idea (as i believe i made it avalibale outside in some commit)

src/sol_imgui/sol_imgui.h
```diff
@@ -2,6 +2,13 @@
 
 namespace sol_ImGui
 {
+
+// Version
+inline std::string GetVersion()
```
WSSDude, line 7: There isnt command for this?

src/sol_imgui/sol_imgui.h
```diff
@@ -2,6 +2,13 @@
 
 namespace sol_ImGui
```
WSSDude, line 3: also, please, document in sol_ImGui readme present in the same folder, it is something i sort of care about as i tried to keep that file as an updated reference of sorts

src/sol_imgui/sol_imgui.h
```diff
@@ -51,6 +60,18 @@ inline bool BeginChild(const std::string& name, float sizeX, float sizeY, bool b
 {
     return ImGui::BeginChild(name.c_str(), {sizeX, sizeY}, border, flags);
 }
+
+// new
```
WSSDude, line 64: i wouldnt personally care about marking the new stuff

WSSDude: reasoning - itis sort of obvious from the history of the changes done tothe file from git itself, IF, i would only marked things which are now deprecated for better visualisation when updating the file later

src/sol_imgui/sol_imgui.h
```diff
@@ -51,6 +60,18 @@ inline bool BeginChild(const std::string& name, float sizeX, float sizeY, bool b
...
+inline bool BeginChild(const std::string& name, float sizeX, float sizeY, int child_flags)
+{
+    return ImGui::BeginChild(name.c_str(), {sizeX, sizeY}, static_cast<ImGuiChildFlags>(child_flags));
+}
+
+inline bool BeginChild(const std::string& name, float sizeX, float sizeY, int child_flags, int window_flags)
```
WSSDude, line 70: please, use int32_t inatead of int

xmake.lua
WSSDude: looking at full changes, you included this only for CET, if ImGui conditionally compiles something by a chance by this, it only does it for files directly contained by the CET and not by the ImGui package

user config gets alao passed into the package

src/sol_imgui/sol_imgui.h
WSSDude: *document the new overloads

src/imgui_impl/win32.h
```diff
@@ -1,40 +1,54 @@
...
+IMGUI_IMPL_API void     ImGui_ImplWin32_Shutdown();
+IMGUI_IMPL_API void     ImGui_ImplWin32_NewFrame(SIZE aOutSize);
 
 // Win32 message handler your application need to call.
-extern IMGUI_IMPL_API LRESULT ImGui_ImplWin32_WndProcHandler(HWND ahWnd, UINT auMsg, WPARAM awParam, LPARAM alParam);
+// - Intentionally commented out in a '#if 0' block to avoid dragging dependencies on <windows.h> from this helper.
+// - You should COPY the line below into your .cpp code to forward declare the function and then you can call it.
+// - Call from your application's message handler. Keep calling your message handler unless this function returns TRUE.
+
+#if 0
```
WSSDude, line 35: just change this to 1 like it was before or remove the ifdef completely please, dont redefine it... i stand behind this, the backends are meant to be copied and customized, they are not *production* ready.

it is fine to change them and implementors *should* change them

src/overlay/widgets/Settings.cpp
```diff
@@ -54,7 +54,7 @@ void Settings::OnUpdate()
         m_madeChanges = false;
         if (ImGui::CollapsingHeader("Patches", ImGuiTreeNodeFlags_DefaultOpen))
         {
-            ImGui::TreePush();
+            ImGui::TreePush((void*)nullptr);
```
WSSDude, line 57: I would just input the actual table constant this is refering to, which is the one in the if

i believe it accepts string, and if it doesnt, just cast that to void*

also, cast using static_cast please, not C style cast.

applies to similar cases

WSSDude: reasoning - anything using nullptr would get treated the same way, this could break the UI (unsure if it really does atm, but it makes this share the ID of the push with other windows/childs/whatev these are in the end...)

it is bug prone if accidentally closed the wrong way

src/d3d12/D3D12.h
WSSDude: as pointed out by another comment, just change it to 1 or remove please

src/d3d12/D3D12_Functions.cpp
```diff
@@ -497,13 +495,13 @@ void D3D12::PrepareUpdate()
 
     for (auto i = 0; i < drawData.CmdListsCount; ++i)
         IM_DELETE(drawData.CmdLists[i]);
-    delete[] drawData.CmdLists;
```
WSSDude, line 500: this is wrong, it leaks memory. revert. both of these lines.

IM_DELETE does not remove all fo the data, only the inner data.

src/d3d12/D3D12_Functions.cpp
```diff
@@ -497,13 +495,13 @@ void D3D12::PrepareUpdate()
...
-    drawData.CmdLists = nullptr;
     drawData.Clear();
 
     drawData = *ImGui::GetDrawData();
 
-    auto** copiedDrawLists = new ImDrawList*[drawData.CmdListsCount];
+    ImVector<ImDrawList*> copiedDrawLists;
```
WSSDude, line 502: this is also wrong, you pass in address on the stack! this is not memory safe at all, you again leak memory i  a very bad way

WSSDude (changes requested):

src/d3d12/D3D12_Functions.cpp
WSSDude: i now understand why you did it i suppose, looking at the line right after...

you are corrupting memory, revert this as a whole, the file shouldnt be changed

src/sol_imgui/sol_imgui.h
dragonzkiller: oh yeah sorry i'll remove this. it's back from when i was doing some initial testing.

xmake.lua
dragonzkiller: moved it to `imgui_user_config.h`

src/sol_imgui/sol_imgui.h
dragonzkiller: the entire rest of the header uses `int` rather than `int32_t`. should i update all of them?

src/d3d12/D3D12_Functions.cpp
```diff
@@ -497,13 +495,13 @@ void D3D12::PrepareUpdate()
...
+    copiedDrawLists.resize(drawData.CmdListsCount);
+
     for (auto i = 0; i < drawData.CmdListsCount; ++i)
         copiedDrawLists[i] = drawData.CmdLists[i]->CloneOutput();
     drawData.CmdLists = copiedDrawLists;
```
WSSDude, line 507: change to use std::move

src/d3d12/D3D12_Functions.cpp
WSSDude: this had changed in the ImGui structure we use apparently, not obvious from changes

src/sol_imgui/sol_imgui.h
WSSDude: i guess you may bulk change them

WSSDude: actually lets not do this in this PR

WSSDude (approved):

PR #977, opened 2024-12-11
Added compatibility with Patch 2.2

alphanin9:
Note: RED4ext.SDK is currently from my branch with 2.2 update, probably should wait until it gets merged into mainline?


WSSDude (approved):
RED4ext probably only blocking factor but otherwise, this works as expected with few test mods, including exEntitySpawner

PR #986, opened 2024-12-14
Add Overlay Binding Notification on Game Start

dragonzkiller:
Took us 4 years, but this commit adds an small notification to the bottom left of the screen with the current bind to remind people what it was set to.

Example shown in Discord: https://discord.com/channels/717692382849663036/795081899756224572/1300137119255232573

### DIFF
--- a/src/overlay/Overlay.cpp
+++ b/src/overlay/Overlay.cpp
@@ -22,6 +22,8 @@ void Overlay::PostInitialize()
         }
 
         m_initialized = true;
+
+        m_startTime = std::chrono::steady_clock::now();
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/overlay/Overlay.cpp
```diff
@@ -195,12 +197,40 @@ void Overlay::Update()
         }
     }
 
+    // Display binding in bottom right corner for 5 seconds
+    if (!Bindings::IsFirstTimeSetup() && !m_bindHintShown)
```
WSSDude, line 201: I would rather do something like in `feature/notifications` branch instead of this, and consider expanding it to sol_ImGui for mods to use.

Maybe even add some setting for the type of spdlog loglevel to notify about and automatically notify when someone logs something. Could be very useful for mods and also internal errors.

The branch needs tweaking, is just a starting point basically reimplementing this PR. It has issues with close button looking weird (probably collision with our style vs. default ImGui style the ImGuiNotify expects...) and is missing license in ThirdParty_LICENSE file.

WSSDude (changes requested):

PR #999, opened 2025-01-26
PDB generation for the release artifacts

alphanin9:
This should help out debugging crash reports in the wild, as most people will not compile CET themselves lol

Not sure about the packaging hack to get .pdb path, is there a more XMakey way?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -50,3 +50,8 @@ jobs:
       with:
         name: ${{ matrix.os }}-${{ matrix.arch }}-${{ matrix.mode }}
         path: package/**
+    
+    - uses: actions/upload-artifact@v4
+      with:
+        name: ${{ matrix.os }}-${{ matrix.arch }}-${{ matrix.mode }}-pdb
+        path: package_pdb/**
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,9 +49,11 @@ jobs:
     - uses: actions/upload-artifact@v4
       with:
         name: ${{ matrix.os }}-${{ matrix.arch }}-${{ matrix.mode }}
-        path: package/**
+        path: |
+          package/**
+          !package/**/*.pdb
     
     - uses: actions/upload-artifact@v4
       with:
         name: ${{ matrix.os }}-${{ matrix.arch }}-${{ matrix.mode }}-pdb
-        path: package_pdb/**
+        path: package/**/*.pdb
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ build/*
 *.asi
 src/CETVersion.h
 package/*
+package_pdb/*
 vs2022/*
 vsxmake*/
 *.pyc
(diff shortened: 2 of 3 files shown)


xmake.lua
```diff
@@ -96,7 +97,16 @@ target("cyber_engine_tweaks")
...
+        local target_file = target:targetfile()
+
+        os.cp(target_file, "package/bin/x64/plugins/")
+
+        os.mkdir("package_pdb/bin/x64")
+
+        os.cp(path.join(
+            path.directory(target_file),
+            path.basename(target_file)..".pdb"
+        ), "package_pdb/bin/x64/plugins/")
```
wopss, line 109: Why copy them to another directory?

.github/workflows/build.yml
```diff
@@ -50,3 +50,8 @@ jobs:
       with:
         name: ${{ matrix.os }}-${{ matrix.arch }}-${{ matrix.mode }}
         path: package/**
+    
+    - uses: actions/upload-artifact@v4
+      with:
+        name: ${{ matrix.os }}-${{ matrix.arch }}-${{ matrix.mode }}-pdb
+        path: package_pdb/**
```
wopss, line 57: Related to https://github.com/maximegmd/CyberEngineTweaks/pull/999/files#r1929751443, this can be done using globs, not separate dirs.

xmake.lua
alphanin9: Was mirroring RED4ext's work to an extent.

.github/workflows/build.yml
wopss: In RED4ext I did this because I wanted to compute the checksum of the ZIP files. But the PDBs are in the same directory as the DLLs.

Here you can change both upload artifacts as below:

```suggestion
    # Upload artifacts
    - uses: actions/upload-artifact@v4
      with:
        name: ${{ matrix.os }}-${{ matrix.arch }}-${{ matrix.mode }}
        path: |
          package/**
          !package/**/*.pdb

    - uses: actions/upload-artifact@v4
      with:
        name: ${{ matrix.os }}-${{ matrix.arch }}-${{ matrix.mode }}-pdb
        path: package/**/*.pdb
```

The only thing I'm not sure about, is if the directory structure will be kept, but I assume it will.

alphanin9: Oh, right - thanks for the info, I'll test that variant when I get to a PC again.

alphanin9: Yeah it does preserve PDB path, you're right - changed it

.gitignore
```diff
@@ -14,6 +14,7 @@ build/*
 *.asi
 src/CETVersion.h
 package/*
+package_pdb/*
```
wopss, line 17:
```suggestion
```

.github/workflows/build.yml
alphanin9: Fixed according to review

xmake.lua
alphanin9: Fixed according to review

maximegmd (approved):

PR #1003, opened 2025-02-18
Fix overlay rendering issues that lead to crashing when other overlays are present

Kaldaien:
I have been getting several reports from Special K users of crashes when using this mod, and finally had some time to investigate.

Primary issues found:

 1. Code to show/hide the hardware cursor does not handle pathological case where something is preventing the cursor's state from changing -- infinite loop.
 2. DisableThreadLibraryCalls (...) used in a DLL that uses MSVCRT managed thread_local storage.
 3. A single command allocator/list was being used for all frame contexts, despite creating per-context allocators.
 4. The game's window title was being used to find the main window, but this is highly volatile; use window class name instead.

### DIFF
--- a/src/d3d12/D3D12.cpp
+++ b/src/d3d12/D3D12.cpp
@@ -9,17 +9,28 @@
 
 void D3D12::SetTrapInputInImGui(const bool acEnabled)
 {
+    // Must have an out-condition to this otherwise infinite loop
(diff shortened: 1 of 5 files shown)


maximegmd:
Thank you very much for the contribution.

I also believe there was a discussion regarding our hooking methodology not handling the SL proxy correctly but did not investigate more.

PR #1004, opened 2025-02-22
Better versioning

### LINKED ISSUE #894: Feature Request: include CET version number within the asi (or somewhere else)
Currently the only way to access a users CET version number outside of the engine is to regex through the log - I have a troubleshooting script which parses log files, core mod versions and game version in an effort to expedite diagnosing mod issues to help users, but getting the CET version requires special handling and due to my crummy regex, is prone to errors. If this is something which could be easily added to the asi or a resource file somewhere that's easily accessed but not easily user manipulated it would be a great time saver for me. 

If this is at all more work than "put into here" for you folks, please just close this issue and ignore and I'll continue to grab from the log file.

WSSDude:
* Version info based off git information
* Add autogenerated resource file so the final DLL has some version info
* Cleaner inclusion of autogenerated files (prevents constant stdafx.h rebuilds and puts resulting files into some more normal directory)
* Fix #894

Untitled

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ package/*
 vs2022/*
 vsxmake*/
 *.pyc
+src/config/*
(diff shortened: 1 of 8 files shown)


maximegmd (approved):
Looks good!

PR #1009, opened 2025-04-08
Add TrapInput script function.

defmys:
- Added TrapInput script function to allow mods to show window with the mouse cursor displayed without opening the CET main overlay.
- Modified D3D12::DelayedSetTrapInputInImGui to properly handle multiple calls from different windows, including CET overlay.

https://github.com/user-attachments/assets/993aa2eb-78ef-42e8-945b-8a4427fb692d

### DIFF
--- a/src/d3d12/D3D12.cpp
+++ b/src/d3d12/D3D12.cpp
@@ -20,8 +20,31 @@ void D3D12::SetTrapInputInImGui(const bool acEnabled)
 
 void D3D12::DelayedSetTrapInputInImGui(const bool acEnabled)
 {
-    m_delayedTrapInputState = acEnabled;
-    m_delayedTrapInput = true;
+    if (acEnabled)
+    {
+        // Trap input if it's the first request.
+        if (m_trapInputStack == 0)
+        {
+            m_delayedTrapInputState = acEnabled;
+            m_delayedTrapInput = true;
+        }
+
+        ++m_trapInputStack;
+    }
+    else
+    {
+        if (m_trapInputStack > 0)
+        {
+            --m_trapInputStack;
+
+            // Disable when the last request is removed.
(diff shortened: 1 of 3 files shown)


maximegmd (approved):
Good addition :)

defmys:
@WSSDude I totally understand your concern and agree that some mods may block the input due to script errors or other reasons. However, most native UI mods I found that support text input accept only English characters by handling keyboard events, but what we need is the ability to type non-English characters using the OS input mothod.

Is it possible to do that with native UI, or perhaps through some other approach? Thanks in advance!

defmys:
Hi @maximegmd @WSSDude , 
Any thoughts on my last reply? Would be great to hear from you!  :)

Rool-IO:
> @WSSDude I totally understand your concern and agree that some mods may block the input due to script errors or other reasons. However, most native UI mods I found that support text input accept only English characters by handling keyboard events, but what we need is the ability to type non-English characters using the OS input mothod.
> 
> Also, I think I didn’t make it clear in the PR description — what we actually need is a text input UI, but it won’t work without proper focus handling. This is essential for typing, especially with OS-level input methods.
> 
> Is it possible to do that with native UI, or perhaps through some other approach? Thanks in advance!

CET Console already allows for typing non English characters in it, not sure what characters you're actually referring to but unless it's relevant to handling mods (I doubt it even remotely is), shouldn't downgrade UX by blocking inputs. Only if user wants that to happen should it do it (kinda like SpecialK lets user decide whether to block types of inputs to a game when its control panel is up).
Just an end-user opinion.

defmys:
@WSSDude Thanks for your reply. It seems Codeware relies on key events to handle text input, which means only English characters are supported. I'll take a look on Discord to see if there's any other aproach. Thanks again for your help!

PR #1034, opened 2025-09-19
Various fixes and bitfield type support

psiberx:
- Fixed `NewProxy` crashes and corrupted lua state in some cases
- Fixed `TweakDBID` ctor crash when a non-null-terminated string is passed
- Fixed `UnknownType` memory leak
- Fixed lua error logging for `Override`s
- Added basic bitfield type support (properties treated as integers)

### DIFF
--- a/src/reverse/BasicTypes.h
+++ b/src/reverse/BasicTypes.h
@@ -132,14 +132,14 @@ struct TweakDBID
 
     TweakDBID(const std::string_view aName)
     {
-        name_hash = RED4ext::CRC32(aName.data(), 0);
+        name_hash = RED4ext::CRC32(reinterpret_cast<const uint8_t*>(aName.data()), aName.size(), 0);
         name_length = static_cast<uint8_t>(aName.size());
         unk5 = unk7 = 0;
     }
 
     TweakDBID(const TweakDBID& aBase, const std::string_view aName)
     {
-        name_hash = RED4ext::CRC32(aName.data(), aBase.name_hash);
+        name_hash = RED4ext::CRC32(reinterpret_cast<const uint8_t*>(aName.data()), aName.size(), aBase.name_hash);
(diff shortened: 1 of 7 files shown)


WSSDude (approved):

src/reverse/BasicTypes.h
```diff
@@ -132,14 +132,14 @@ struct TweakDBID
 
     TweakDBID(const std::string_view aName)
     {
-        name_hash = RED4ext::CRC32(aName.data(), 0);
+        name_hash = RED4ext::CRC32(reinterpret_cast<const uint8_t*>(aName.data()), aName.size(), 0);
```
WSSDude, line 135: I just had a better look now because of what was raised on Discord, and if this is done only because some characters in string may get interpreted as negative value and go out of the array somehow, think the fix should be in the RED4ext.SDK and not here

WSSDude (changes requested):

src/reverse/BasicTypes.h
WSSDude: Ok, there is an issue of no null terminator actually in some instances from discussion so this comment is invalid.

src/reverse/Converter.h
```diff
@@ -183,6 +183,75 @@ struct EnumConverter : LuaRED<Enum, "Enum">
...
+    {
+        uint64_t value = 0;
+
+        if (aObject.get_type() == sol::type::number)
+        {
+            value = aObject.as<uint32_t>();
+        }
+        else if (IsLuaCData(aObject))
+        {
+            sol::state_view v(aObject.lua_state());
```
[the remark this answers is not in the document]
WSSDude: Looking at other converters likely fine...

WSSDude (approved):

psiberx:
I think that's all for now :)

PR #1046, opened 2025-12-28
Enable Docking

notaspirit:
# Enable Docking

Added:
* Docking and Docking Over Viewport (so windows can be docked to the viewport, not just other windows) 

Changed:
* ImGui now only gets passed the input when the overlay is open, the trap continues to consume all events after

Removed:
* No Mouse Course Change Flag which was causing incorrect mouse cursor state

### DIFF
--- a/src/d3d12/D3D12.cpp
+++ b/src/d3d12/D3D12.cpp
@@ -38,7 +38,8 @@ LRESULT D3D12::OnWndProc(HWND ahWnd, UINT auMsg, WPARAM awParam, LPARAM alParam)
 
         if (d3d12.m_trapInputInImGui) // TODO: look into io.WantCaptureMouse and io.WantCaptureKeyboard
         {
-            ImGui_ImplWin32_WndProcHandler(ahWnd, auMsg, awParam, alParam);
+            if (const LRESULT res = ImGui_ImplWin32_WndProcHandler(ahWnd, auMsg, awParam, alParam))
+                return res;
 
             // ignore mouse & keyboard events
             if ((auMsg >= WM_MOUSEFIRST && auMsg <= WM_MOUSELAST) || (auMsg >= WM_KEYFIRST && auMsg <= WM_KEYLAST))
--- a/src/d3d12/D3D12.cpp
+++ b/src/d3d12/D3D12.cpp
@@ -30,9 +30,6 @@ LRESULT D3D12::OnWndProc(HWND ahWnd, UINT auMsg, WPARAM awParam, LPARAM alParam)
 
     if (d3d12.IsInitialized())
     {
-        if (const auto res = ImGui_ImplWin32_WndProcHandler(ahWnd, auMsg, awParam, alParam))
-            return res;
-
         if (d3d12.m_delayedTrapInput)
         {
             d3d12.SetTrapInputInImGui(m_delayedTrapInputState);
@@ -41,6 +38,8 @@ LRESULT D3D12::OnWndProc(HWND ahWnd, UINT auMsg, WPARAM awParam, LPARAM alParam)
 
         if (d3d12.m_trapInputInImGui) // TODO: look into io.WantCaptureMouse and io.WantCaptureKeyboard
         {
+            ImGui_ImplWin32_WndProcHandler(ahWnd, auMsg, awParam, alParam);
+
             // ignore mouse & keyboard events
             if ((auMsg >= WM_MOUSEFIRST && auMsg <= WM_MOUSELAST) || (auMsg >= WM_KEYFIRST && auMsg <= WM_KEYLAST))
                 return 1;
--- a/src/d3d12/D3D12_Functions.cpp
+++ b/src/d3d12/D3D12_Functions.cpp
@@ -313,7 +313,7 @@ bool D3D12::InitializeImGui(size_t aBuffersCounts)
     ImGui::GetStyle() = m_styleReference;
     ImGui::GetStyle().ScaleAllSizes(scaleFromReference);
 
-    ImGui::GetIO().ConfigFlags |= ImGuiConfigFlags_NoMouseCursorChange; // Do not modify cursor from ImGui backend
+    ImGui::GetIO().ConfigFlags |= ImGuiConfigFlags_DockingEnable;
 
     if (!ImGui_ImplWin32_Init(m_window.GetWindow()))
     {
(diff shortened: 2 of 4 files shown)


WSSDude:
If this is a draft mark it as one. Otherwise may have a look later (unsure if today but should be able to tomorrow).

WSSDude:
Saying that as you seem to be pushing still unsure if this is final

notaspirit:
yeah it's done now, already marked it as a non draft.

src/d3d12/D3D12.cpp
```diff
@@ -41,6 +38,8 @@ LRESULT D3D12::OnWndProc(HWND ahWnd, UINT auMsg, WPARAM awParam, LPARAM alParam)
 
         if (d3d12.m_trapInputInImGui) // TODO: look into io.WantCaptureMouse and io.WantCaptureKeyboard
         {
+            ImGui_ImplWin32_WndProcHandler(ahWnd, auMsg, awParam, alParam);
```
dragonzkiller: Why move this down here and ignore the result?

src/d3d12/D3D12_Functions.cpp
```diff
@@ -313,7 +313,7 @@ bool D3D12::InitializeImGui(size_t aBuffersCounts)
     ImGui::GetStyle() = m_styleReference;
     ImGui::GetStyle().ScaleAllSizes(scaleFromReference);
 
-    ImGui::GetIO().ConfigFlags |= ImGuiConfigFlags_NoMouseCursorChange; // Do not modify cursor from ImGui backend
+    ImGui::GetIO().ConfigFlags |= ImGuiConfigFlags_DockingEnable;
```
dragonzkiller, line 316: Why remove the other flag? Does it prevent docking in some way?

WSSDude: I believe it should solve the cursor not changing based on if you are over the text, selecting or draging n such but unsure. Also unsure why it was even added now overall but that is one of the things I wanted to check later.

WSSDude: Leave this open for now, would likely also open a thread

src/d3d12/D3D12.cpp
notaspirit: The not handling of the result is an oversight on my part, will fix that.

Initially I moved it here in this pr because I noticed you can drag windows around that are always drawn even when the overlay is closed which I assumed was due to the Dockingspace over viewport but after now testing it again it turns out it was an issue even before that. 
Moving it to only handle input then the overlay is open also fixes the issue of the os cursor not hiding, which is the reason the NoMouseCursorChange was needed in the first place.
If you prefer I can move this (and the flag removal) out to a separate pr.

src/d3d12/D3D12_Functions.cpp
notaspirit: With the change in when imgui can handle the input it is no longer needed as it was used to prevent the os cursor from being visible when the overlay is closed.

src/d3d12/D3D12.cpp
WSSDude: Nah is good likely but also sth I want to overall check with the rest soon tm

src/overlay/Overlay.cpp
```diff
@@ -61,6 +61,8 @@ void Overlay::Update()
     if (!m_initialized)
         return;
 
+    ImGui::DockSpaceOverViewport(0, ImGui::GetMainViewport(), ImGuiDockNodeFlags_PassthruCentralNode);
```
WSSDude, line 64: Expose related docking functions to Lua also please, and update the sol_ImGui readme. Also, this will likely allow toolbar and few other things to get docked as it is now which I dont really like personally. It would be better if the toolbar was docked by default in some way and prevent manipulation, sth I wanted to do with the docking (and reason it was not added still kinda, wanted to do it more properly or how to call it).

Toolbar shouldnt be a floating window with the docking

WSSDude (changes requested):

src/d3d12/D3D12_Functions.cpp
notaspirit: and yeah removing that flag restores the cursor changes by imgui with e.g. the text cursor or the drag/arrow cursor when resizing etc.

src/d3d12/D3D12.cpp
WSSDude: One thing that isnt good though is the return, that has to happen to block input passthrough to the game. Some change that we did in imgui_impl_win32 likely causes some issues due to which you had to remove the return.

But yeah have to overall check as mentioned

src/overlay/Overlay.cpp
notaspirit: For the toolbar it should be a main menu bar then which is always at the top of the viewport, and that is possible regardless of docking.

WSSDude: Yeah also had that idea overall, but just the menu bar didnt quite looked right. Think it would be better to split the main viewport into 2 docking zones and just dock the toolbar into the top zone. This requires a bit more of a setup to split the docking zones and also make the top docking zone fixed, along with programatically docking the toolbar into the top zone

WSSDude: If you are unsure how to do it may dig up piece of code from the legendary 2.0 had this working (and imo it looked just better than a simple menu)

notaspirit: by docking related methods are you referring to the [DockBuider API](<https://github.com/ocornut/imgui/wiki/Docking#programmatically-setting-up-docking-layout-dockbuider-api>)?
Not sure what a reasonable approach would be for implementing that, might need it's own event since it seems to need to be done before the main docking space draws from just looking at the example.

notaspirit: Just a visual example of what kind of menu you had in mind would be helpful too, pretty sure the main menu bar can be heavily customized too and is by default excluded from docking, immovable etc. so it might be better to use it rather than a manual window.

WSSDude: Hm, possible, hadnt truly tried to customize it. Like if you have some idea feel free to try to make sth and post some screen how it looks.

And yes, youd want to make a new event onDockingSetup or something like that and call it once after ImGui initialization.

WSSDude: Note that "once" may be multiple times, you want to reset the flag when the ImGui is reinitialized during reset of state.

WSSDude: Actually I thought now - this may overall baloon a bit with these additions, we may do it separately (Lua API).

But Id like to resolve the toolbar at least with this.

maximegmd (approved):

PR #1049, opened 2026-01-04
Improved cet window

notaspirit:
# Improved CET window

CET window toggles and reload mods button are now in the main menu bar (always at the top of the screen).
Widgets are additionally in a submenu now.


notaspirit:
It's now looking like this, where it keeps the old layout but still in the menu bar
image

PR #1055, opened 2026-04-19
fix: Crash when attempting to construct abstract class

budjb:
This small fix address situations where attempts to construct an abstract class results in a crash.

Currently, the code explicitly returns `nullptr` when constructing the class instance if the class has been marked as abstract, but the subsequent handle conversion logic does not check for the null value before attempting to convert it to a `Handle`. This guard ensures that the `nil` type is returned to Lua, resolving the crash scenario.

To reproduce, simply attempt to create a new instance of an abstract class like so:

```lua
MyAbstractType.new()
```


maximegmd (approved):
lgtm

